PR #108, opened 2026-03-16
feat: add MemoryClient for control plane + data plane passthrough

Hweinstock:
## Description

Add MemoryClient for Bedrock AgentCore Memory, covering both control plane and data plane operations.

Uses a Proxy with [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) to expose a curated set of AWS SDK methods as direct passthroughs on the client.

`.memory(memoryId)` returns a scoped ScopedMemory that auto-injects the memoryId into every call, reducing repetition when operating on a single memory.

Convenience methods for common workflows:
- `createOrGetMemory` — idempotent create (falls back to get on conflict)
- `createMemoryAndWait` / `deleteMemoryAndWait` — polling wrappers
- `waitForMemories` — polls until LTM extraction produces records
- `getLastKTurns` — retrieves the last K conversation turns grouped by USER boundaries
- `listBranches` — aggregates branch info from events

Core design decisions vs Python SDK:
- Single class instead of Python's 3 (MemoryClient, MemoryControlPlaneClient, MemorySessionManager)
- No strategy helpers — SDK's UpdateMemoryCommand already has full typed configs
- No response normalization or wrapper models — TS SDK types are canonical
- Strands integration will not live here and will be decoupled from core memory functionality

NOTE: Not yet published into the vended SDK, no entry in the package.json exports map. Will be added once integration tests are in place.

## Testing

Unit tests use dependency injection (injectable SDK clients via `MemoryClientConfig`) with simple fakes to avoid mocks. Tests verify:
- Passthrough routing: every data plane and control plane method is exposed as a function on the real client
- ScopedMemory: scoped methods are callable
- Convenience methods: createOrGetMemory conflict handling, deleteMemoryAndWait polling, getLastKTurns turn boundary parsing, listBranches aggregation

## Type of Change

New feature

## Checklist

https://github.com/aws/bedrock-agentcore-sdk-typescript/issues/125

### DIFF
--- a/src/_utils/polling.ts
+++ b/src/_utils/polling.ts
@@ -1,0 +1,29 @@
+/**
+ * Poll a condition function until it returns `true` or the timeout expires.
+ *
+ * @param condition - Async function that returns `true` when done. Throwing is treated as "not done yet" when `swallowErrors` is true.
+ * @param opts - Polling options
+ * @returns `true` if the condition was met, `false` if timed out (only when `timeoutErrorMessage` is not set)
+ * @throws Error with `timeoutErrorMessage` if provided and timeout expires
+ */
+export async function pollUntil(
+  condition: () => Promise<boolean>,
+  opts: {
+    maxWaitSeconds: number
+    pollIntervalMs: number
+    timeoutErrorMessage?: string
+    shouldSwallowError?: (err: unknown) => boolean
+  },
+): Promise<boolean> {
+  const deadline = Date.now() + opts.maxWaitSeconds * 1000
+  while (Date.now() < deadline) {
+    try {
+      if (await condition()) return true
+    } catch (err) {
+      if (!opts.shouldSwallowError?.(err)) throw err
+    }
+    await new Promise((resolve) => globalThis.setTimeout(resolve, opts.pollIntervalMs))
+  }
+  if (opts.timeoutErrorMessage) throw new Error(opts.timeoutErrorMessage)
+  return false
+}
--- a/src/memory/__tests__/client.test.ts
+++ b/src/memory/__tests__/client.test.ts
@@ -1,0 +1,172 @@
+import { describe, it, expect } from 'vitest'
+import type { BedrockAgentCore } from '@aws-sdk/client-bedrock-agentcore'
+import type { BedrockAgentCoreControl } from '@aws-sdk/client-bedrock-agentcore-control'
+import { MemoryClient } from '../client.js'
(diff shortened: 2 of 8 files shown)


src/memory/client.ts
```diff
@@ -0,0 +1,284 @@
...
+          name,
+          rootEventId: event.branch?.rootEventId,
+          eventCount: 1,
+        })
+      }
+    }
+    return Array.from(branches.values())
+  }
+
+  memory(memoryId: string): ScopedMemory {
```
jariy17, line 234: I like this design, but I dont understand the benefits of having a ScopeMemory. Could you explain why a customer just provide the memoryId themselves each time they call createEvent and similar functions?

Hweinstock: mostly verbosity, there is no reason they can't manually provide an ID each time. I saw it as concise way to combine a memory-scoped client, and a raw SDK client into a single interface. The same functionality is split between clients in the existing python implementation. 

Happy to leave it out, if we think its awkward/wont be used.

jariy17: lets leave it out for now

src/_utils/polling.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Poll a condition function until it returns `true` or the timeout expires.
+ *
+ * @param condition - Async function that returns `true` when done. Throwing is treated as "not done yet" when `swallowErrors` is true.
+ * @param opts - Polling options
+ * @returns `true` if the condition was met, `false` if timed out (only when `timeoutErrorMessage` is not set)
+ * @throws Error with `timeoutErrorMessage` if provided and timeout expires
+ */
+export async function pollUntil(
```
jesseturner21, line 9: nice!

src/memory/__tests__/client.test.ts
```diff
@@ -0,0 +1,172 @@
...
+  return new Proxy({} as Record<string, unknown>, {
+    get: (_, method) => overrides[method as string] ?? (() => Promise.resolve({})),
+  })
+}
+
+function fakeDataPlane(overrides: Record<string, (input: unknown) => unknown>): Record<string, unknown> {
+  return new Proxy({} as Record<string, unknown>, {
+    get: (_, method) => overrides[method as string] ?? (() => Promise.resolve({})),
+  })
+}
```
jesseturner21, line 17: SUPER NIT: is there any reason these are diff, they are both the same implementation just diff names. 

It could be something like `fakeClient(....)`

PR #139, opened 2026-04-16
chore: merge PR #108 MemoryClient into feat/agentcore-memory

jesseturner21:
## Description

Merges [PR #108](https://github.com/aws/bedrock-agentcore-sdk-typescript/pull/108) (`feat: add MemoryClient for control plane + data plane passthrough`) into the `feat/agentcore-memory` feature branch so the Strands plugin work can build on top of the new `MemoryClient`.

This is a merge-forward only — no additional changes on top of PR #108's commit (`18dd5ee`).

## Related Issues

- Upstream PR: #108

## Type of Change

Other: integration merge into feature branch

## Testing

- [x] `npm run build` passes
- [x] `npm test` passes (401 tests, 19 files)

## Checklist
- [x] Tests from PR #108 are included and passing

----

### DIFF
--- a/src/_utils/__tests__/pagination.test.ts
+++ b/src/_utils/__tests__/pagination.test.ts
@@ -1,0 +1,50 @@
+import { describe, it, expect } from 'vitest'
+import { paginateAll } from '../../_utils/pagination.js'
+
+describe('paginateAll', () => {
+  it('should return all items from a single page', async () => {
+    const pages = [{ items: [1, 2, 3], nextToken: undefined as string | undefined }]
+    let callCount = 0
+
+    const result = await paginateAll(
+      () => Promise.resolve(pages[callCount++]!),
+      (page) => page.items,
+      (page) => page.nextToken,
+    )
+
+    expect(result).toEqual([1, 2, 3])
+    expect(callCount).toBe(1)
+  })
+
+  it('should accumulate items across multiple pages', async () => {
+    const pages = [
+      { items: [1, 2], nextToken: 'tok1' as string | undefined },
+      { items: [3, 4], nextToken: 'tok2' as string | undefined },
+      { items: [5], nextToken: undefined as string | undefined },
+    ]
+    let callCount = 0
+    const tokensReceived: (string | undefined)[] = []
+
+    const result = await paginateAll(
+      (nextToken) => {
+        tokensReceived.push(nextToken)
+        return Promise.resolve(pages[callCount++]!)
+      },
+      (page) => page.items,
+      (page) => page.nextToken,
+    )
+
+    expect(result).toEqual([1, 2, 3, 4, 5])
+    expect(tokensReceived).toEqual([undefined, 'tok1', 'tok2'])
+  })
+
+  it('should return empty array when page has no items', async () => {
+    const result = await paginateAll(
+      () => Promise.resolve({ items: undefined as number[] | undefined, nextToken: undefined }),
+      (page) => page.items,
+      (page) => page.nextToken,
+    )
+
+    expect(result).toEqual([])
+  })
+})
--- a/src/_utils/__tests__/pagination.test.ts
+++ b/src/_utils/__tests__/pagination.test.ts
@@ -9,7 +9,7 @@ describe('paginateAll', () => {
     const result = await paginateAll(
       () => Promise.resolve(pages[callCount++]!),
       (page) => page.items,
-      (page) => page.nextToken,
+      (page) => page.nextToken
     )
 
     expect(result).toEqual([1, 2, 3])
@@ -31,7 +31,7 @@ describe('paginateAll', () => {
         return Promise.resolve(pages[callCount++]!)
       },
       (page) => page.items,
-      (page) => page.nextToken,
+      (page) => page.nextToken
     )
 
     expect(result).toEqual([1, 2, 3, 4, 5])
@@ -42,7 +42,7 @@ describe('paginateAll', () => {
     const result = await paginateAll(
       () => Promise.resolve({ items: undefined as number[] | undefined, nextToken: undefined }),
       (page) => page.items,
-      (page) => page.nextToken,
+      (page) => page.nextToken
     )
 
     expect(result).toEqual([])
--- a/src/_utils/polling.ts
+++ b/src/_utils/polling.ts
@@ -1,0 +1,29 @@
+/**
+ * Poll a condition function until it returns `true` or the timeout expires.
+ *
+ * @param condition - Async function that returns `true` when done. Throwing is treated as "not done yet" when `swallowErrors` is true.
+ * @param opts - Polling options
+ * @returns `true` if the condition was met, `false` if timed out (only when `timeoutErrorMessage` is not set)
+ * @throws Error with `timeoutErrorMessage` if provided and timeout expires
+ */
+export async function pollUntil(
+  condition: () => Promise<boolean>,
+  opts: {
+    maxWaitSeconds: number
+    pollIntervalMs: number
+    timeoutErrorMessage?: string
+    shouldSwallowError?: (err: unknown) => boolean
+  },
+): Promise<boolean> {
+  const deadline = Date.now() + opts.maxWaitSeconds * 1000
+  while (Date.now() < deadline) {
+    try {
+      if (await condition()) return true
+    } catch (err) {
+      if (!opts.shouldSwallowError?.(err)) throw err
+    }
+    await new Promise((resolve) => globalThis.setTimeout(resolve, opts.pollIntervalMs))
+  }
+  if (opts.timeoutErrorMessage) throw new Error(opts.timeoutErrorMessage)
+  return false
+}
--- a/src/_utils/polling.ts
+++ b/src/_utils/polling.ts
@@ -13,7 +13,7 @@ export async function pollUntil(
     pollIntervalMs: number
     timeoutErrorMessage?: string
     shouldSwallowError?: (err: unknown) => boolean
-  },
+  }
 ): Promise<boolean> {
   const deadline = Date.now() + opts.maxWaitSeconds * 1000
   while (Date.now() < deadline) {
--- a/src/memory/client.ts
+++ b/src/memory/client.ts
@@ -1,7 +1,4 @@
-import {
-  BedrockAgentCore,
-  type Conversational,
-} from '@aws-sdk/client-bedrock-agentcore'
+import { BedrockAgentCore, type Conversational } from '@aws-sdk/client-bedrock-agentcore'
 import {
   BedrockAgentCoreControl,
   type CreateMemoryCommandInput,
@@ -10,7 +7,6 @@ import {
 } from '@aws-sdk/client-bedrock-agentcore-control'
 import type {
   MemoryClientConfig,
-  ScopedMemory,
   WaitOptions,
   WaitForMemoriesParams,
   GetLastKTurnsParams,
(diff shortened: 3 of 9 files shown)


src/memory/client.ts
```diff
@@ -0,0 +1,244 @@
...
+      },
+      {
+        maxWaitSeconds: params.maxWaitSeconds ?? 180,
+        pollIntervalMs: params.pollIntervalMs ?? 15_000,
+        shouldSwallowError: () => true,
+      }
+    )
+  }
+
+  async getLastKTurns(params: GetLastKTurnsParams): Promise<Conversational[][]> {
```
aidandaly24, line 147: This method returns the **first** K turns of a session, not the last K. The name promises "last K", so this is a correctness bug that will misroute any caller using it for warm-start context, recent-history injection, or tail slicing.

The loop pages forward from the oldest event and stops once `turns.length >= params.k`:

```ts
while (turns.length < params.k) {
  const response = await this.dataPlane.listEvents({ ...listParams, nextToken } ...)
  // appends turns in forward order, stops at K
}
```

On a session with 10 turns and `k=3`, this returns turns 1–3, not 8–10.

**Suggested fix (option A — preferred, if the service supports it):** Pass a reverse-order flag to `listEvents` so the first page contains the newest events, then collect K and reverse before returning.

**Suggested fix (option B — if reverse-order isn't supported):** Page through the whole session via `paginateAll`, then slice the tail:

```ts
const allEvents = await paginateAll(
  (nextToken) => this.dataPlane.listEvents({ ...listParams, nextToken } as ...),
  (page) => page.events,
  (page) => page.nextToken,
)
const allTurns = groupIntoTurns(allEvents)
return allTurns.slice(-params.k)
```

Option B is O(n) in session length — for very long sessions it may be worth a TODO to add a service-side tail API.

Also worth adding a unit test that asserts turn ordering: create a session with turns 1–10, request `k=3`, assert the result is turns 8–10.

src/memory/client.ts
```diff
@@ -0,0 +1,244 @@
...
+
+function hasName(err: unknown, name: string): boolean {
+  return typeof err === 'object' && err !== null && 'name' in err && (err as { name: unknown }).name === name
+}
+
+async function findMemoryIdByName(controlPlane: BedrockAgentCoreControl, name: string): Promise<string | undefined> {
+  let nextToken: string | undefined
+  do {
+    const resp = await controlPlane.listMemories({ nextToken })
+    const match = resp.memories?.find((m) => m.id?.startsWith(`${name}-`) || m.id === name)
```
aidandaly24, line 237: The prefix-match heuristic has a name-collision bug: if a customer has two memories named `prod` and `prod-east`, a `createOrGetMemory({ name: 'prod' })` call may match `prod-east-<suffix>` depending on `listMemories` page ordering.

This is most dangerous because `createOrGetMemory` is typically called once at agent startup, so a silent mismatch routes an entire session's events into the wrong memory with no error surface.

**Suggested fix:** Match on the `name` field directly (the `listMemories` response almost certainly includes it):

```ts
const match = resp.memories?.find((m) => m.name === name)
```

If the `name` field isn't available in the summary and requires an additional `getMemory` call per candidate, at minimum tighten the heuristic and document it:

```ts
// Match requires an exact `<name>-<suffix>` pattern, not just any prefix match.
const SUFFIX_PATTERN = /^[A-Za-z0-9]+$/
const match = resp.memories?.find((m) => {
  if (m.id === name) return true
  const suffix = m.id?.slice(name.length + 1)
  return m.id?.startsWith(`${name}-`) && suffix && SUFFIX_PATTERN.test(suffix)
})
```

And add a JSDoc on `createOrGetMemory` steering customers to cache the returned `memoryId` in production rather than relying on name-based lookup.

src/_utils/__tests__/pagination.test.ts
```diff
@@ -0,0 +1,50 @@
+import { describe, it, expect } from 'vitest'
```
aidandaly24, line 1: **File:** `package.json`
**Line:** 12–57 (the `exports` map, after the `./identity` entry)

**Comment:**

Without this, the clean import path fails:

```ts
import { MemoryClient } from 'bedrock-agentcore/memory'  // ❌ module not found
```

Blocks CLI templates (Workstream A6/A7) and the `AgentCoreMemoryPlugin`'s consumption path. This is task A3 from the plan and still needs to happen in this PR (the PR description mentions it as a known gap).

**Suggested fix:** Add the entry right after `./identity`:

```json
"./identity": {
  "import": "./dist/src/identity/index.js",
  "types": "./dist/src/identity/index.d.ts"
},
"./memory": {
  "import": "./dist/src/memory/index.js",
  "types": "./dist/src/memory/index.d.ts"
},
```

src/memory/client.ts
```diff
@@ -0,0 +1,244 @@
...
+          memoryId: params.memoryId,
+          namespace: params.namespace,
+          searchCriteria: { searchQuery: params.testQuery ?? 'test' },
+        })
+        return (resp.memoryRecordSummaries?.length ?? 0) > 0
+      },
+      {
+        maxWaitSeconds: params.maxWaitSeconds ?? 180,
+        pollIntervalMs: params.pollIntervalMs ?? 15_000,
+        shouldSwallowError: () => true,
```
aidandaly24, line 142: This eats `AccessDeniedException`, `ResourceNotFoundException`, `ValidationException`, service outages — everything — and returns `false` on timeout with zero indication of why. A caller with a misconfigured `memoryId` or missing IAM permissions will poll for 180 seconds and then get `false`, then call `retrieveMemoryRecords` on the same bad config and get an empty result, assuming "no records yet" when the real failure is auth.

**Suggested fix:** Scope error swallowing to transient-only errors:

```ts
async waitForMemories(params: WaitForMemoriesParams): Promise<boolean> {
  return pollUntil(
    async () => {
      const resp = await this.dataPlane.retrieveMemoryRecords({
        memoryId: params.memoryId,
        namespace: params.namespace,
        searchCriteria: { searchQuery: params.testQuery ?? 'test' },
      })
      return (resp.memoryRecordSummaries?.length ?? 0) > 0
    },
    {
      maxWaitSeconds: params.maxWaitSeconds ?? 180,
      pollIntervalMs: params.pollIntervalMs ?? 15_000,
      shouldSwallowError: (err) => {
        const name = (err as { name?: string })?.name
        return name === 'ThrottlingException' || name === 'ServiceUnavailableException' || name === 'InternalServerException'
      },
    }
  )
}
```

src/_utils/polling.ts
```diff
@@ -0,0 +1,29 @@
...
+): Promise<boolean> {
+  const deadline = Date.now() + opts.maxWaitSeconds * 1000
+  while (Date.now() < deadline) {
+    try {
+      if (await condition()) return true
+    } catch (err) {
+      if (!opts.shouldSwallowError?.(err)) throw err
+    }
+    await new Promise((resolve) => globalThis.setTimeout(resolve, opts.pollIntervalMs))
+  }
```
aidandaly24, line 26: Two issues:

1. **Wasted sleep after deadline.** On the final failed iteration, we sleep the full `pollIntervalMs` (15 s for `waitForMemories`) even though the deadline has already passed.
2. **`maxWaitSeconds: 0` never runs the condition.** The deadline check fires before the first condition call, so a zero-timeout configuration silently returns `false` without any attempt.

**Suggested fix:** Check the deadline after the sleep, and run at least one attempt:

```ts
export async function pollUntil(
  condition: () => Promise<boolean>,
  opts: {
    maxWaitSeconds: number
    pollIntervalMs: number
    timeoutErrorMessage?: string
    shouldSwallowError?: (err: unknown) => boolean
  }
): Promise<boolean> {
  const deadline = Date.now() + opts.maxWaitSeconds * 1000
  while (true) {
    try {
      if (await condition()) return true
    } catch (err) {
      if (!opts.shouldSwallowError?.(err)) throw err
    }
    if (Date.now() + opts.pollIntervalMs > deadline) break
    await new Promise((resolve) => globalThis.setTimeout(resolve, opts.pollIntervalMs))
  }
  if (opts.timeoutErrorMessage) throw new Error(opts.timeoutErrorMessage)
  return false
}
```

src/memory/types.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export type DataPlaneMethods = Pick<BedrockAgentCore, (typeof DATA_PLANE_METHODS)[number]>
+export type ControlPlaneMethods = Pick<BedrockAgentCoreControl, (typeof CONTROL_PLANE_METHODS)[number]>
+
+export interface MemoryClientConfig {
+  region?: string
+  credentialsProvider?: AwsCredentialIdentityProvider
+  dataPlaneClient?: BedrockAgentCore
+  controlPlaneClient?: BedrockAgentCoreControl
+}
```
aidandaly24, line 39: The upcoming `AgentCoreMemoryPlugin` will batch-flush N parallel `createEvent` calls per turn. Under load, throttle cascades are realistic. Today there's no way to configure `maxAttempts` or `retryMode` on the SDK clients constructed internally — customers must either live with the SDK defaults or build a full client themselves and inject via `dataPlaneClient` / `controlPlaneClient`.

**Suggested fix:** Pass retry options through to the underlying clients:

```ts
export interface MemoryClientConfig {
  region?: string
  credentials?: AwsCredentialIdentityProvider  // renamed per comment #7
  maxAttempts?: number
  retryMode?: 'standard' | 'adaptive'
  dataPlaneClient?: BedrockAgentCore
  controlPlaneClient?: BedrockAgentCoreControl
}
```

And in the constructor (`client.ts:40`):

```ts
const clientConfig = {
  region,
  ...(config?.credentials && { credentials: config.credentials }),
  ...(config?.maxAttempts && { maxAttempts: config.maxAttempts }),
  ...(config?.retryMode && { retryMode: config.retryMode }),
}
```

aidandaly24: A customer familiar with `@aws-sdk/credential-providers` will expect the field to be named `credentials` (that's the name on every AWS SDK v3 client config). The internal mapping at `client.ts:41` (`credentials: config.credentialsProvider`) transparently handles it, but the mismatch creates copy-paste friction.

**Suggested fix:** Rename `credentialsProvider` → `credentials` in `MemoryClientConfig`, and update the destructure in `client.ts:42`:

```ts
const clientConfig = {
  region,
  ...(config?.credentials && { credentials: config.credentials }),
}
```

If the provider-vs-static distinction is important to preserve semantically, document it in the JSDoc with an example:

```ts
/** Credential provider function (e.g. fromSSO, fromTemporaryCredentials). */
credentials?: AwsCredentialIdentityProvider
```

src/memory/client.ts
```diff
@@ -0,0 +1,244 @@
...
+      },
+      {
+        maxWaitSeconds: opts?.maxWaitSeconds ?? 300,
+        pollIntervalMs: opts?.pollIntervalMs ?? 10_000,
+        timeoutErrorMessage: `Memory ${memoryId} was not deleted within ${opts?.maxWaitSeconds ?? 300}s`,
+      }
+    )
+  }
+
+  async waitForMemories(params: WaitForMemoriesParams): Promise<boolean> {
```
aidandaly24, line 129: A customer writing `await client.waitForMemories(...)` without checking the return value will call `retrieveMemoryRecords` afterwards and get an empty result, concluding "no LTM records yet" when the real outcome was a timeout.

**Suggested fix:** Add a `@returns` tag to the JSDoc, and consider a `throwOnTimeout` option:

```ts
/**
 * Poll the LTM retrieval endpoint until at least one record is returned.
 *
 * @returns `true` if records appeared within the timeout, `false` otherwise.
 *          Never throws on timeout — check the return value or pass
 *          `throwOnTimeout: true` to get an exception instead.
 */
async waitForMemories(params: WaitForMemoriesParams): Promise<boolean> {
  // ...
}
```

Same applies to `pollUntil`'s silent-false path — document the contract clearly in `src/_utils/polling.ts:7`.

src/memory/client.ts
```diff
@@ -0,0 +1,244 @@
...
+  GetLastKTurnsParams,
+  BranchInfo,
+  DataPlaneMethods,
+  ControlPlaneMethods,
+} from './types.js'
+import { DATA_PLANE_METHODS, CONTROL_PLANE_METHODS } from './types.js'
+import { paginateAll, DEFAULT_PAGE_SIZE } from '../_utils/pagination.js'
+import { pollUntil } from '../_utils/polling.js'
+
+const DEFAULT_REGION = 'us-west-2'
```
aidandaly24, line 21: `RuntimeClient` (`src/runtime/client.ts:55-62`) and `IdentityClient` (`src/identity/client.ts:28-36`) both throw when no region is configured. `Browser` and `CodeInterpreter` silently default like this one, so there's already inconsistency in the repo.

I think this is the better way to have it

src/memory/client.ts
jesseturner21: good call am adding

PR #142, opened 2026-04-23
feat(memory): AgentCoreMemory plugin for Strands SDK

aidandaly24:
## Description

Implements the `AgentCoreMemory` plugin for the Strands TypeScript SDK, providing LTM extraction (store conversation messages as memory events) and LTM injection (retrieve insights into the agent's system prompt). This is the TypeScript equivalent of the Python SDK's memory integration, designed from the ground up to avoid the accumulation bug documented in [#420](https://github.com/aws/bedrock-agentcore-sdk-python/issues/420).

### What's included

- **`AgentCoreMemory`** — Strands Plugin (SDK >=0.7.0) with:
  - **Extraction pipeline**: `MessageAddedEvent` → buffer messages → batched flush via `createEvent()` with `Promise.allSettled` + `clientToken` idempotent retries
  - **Injection pipeline**: `BeforeInvocationEvent` → strip stale `<agentcore_memory>` sentinel → retrieve fresh LTM per namespace in parallel → format → inject into `systemPrompt`
  - **`search_memory` tool**: opt-in via `searchTool: true`, registered through `getTools()`
  - **Multi-agent helpers**: `withActor(actorId)`, `withMetadataProvider(fn)`
  - **Lifecycle APIs**: `flush()` and `shutdown()` for graceful drains
  - **Template placeholders**: `{actorId}` / `{sessionId}` resolved at construct time
  - **Constructor validation**: fail-fast on invalid extraction config (batchSize, batchTimeoutMs, flushTimeoutMs, maxDrainIterations)
  - **`onDroppedEvents` callback**: structured drop notifications for retry-failed / timeout / post-shutdown / max-drain-iterations
  - **`AsyncBatcher` utility** (internal): generic buffer/batch/retry/timeout/drain pipeline extracted from the plugin
- **`createSearchMemoryTool`** — standalone tool factory for advanced composability
- **143 unit tests** + integration test scaffold
- **Export path**: `bedrock-agentcore/experimental/memory/strands`

### Key design decisions

- **Extraction and injection are decoupled** — customers opt into each independently
- **Ephemeral system prompt injection** — strip and re-inject each turn, never mutate persisted messages (fixes the Python SDK's O(N×K) accumulation bug)
- **Batching on by default** — `batchSize: 10`, `batchTimeoutMs: 5000ms`. Note: `batchSize` controls *when* the buffer drains, not how many messages are sent per API call — `createEvent` is one-call-per-message regardless of `batchSize` (documented in TSDoc).
- **`extraction:` / `injection:` naming** — replaces `storeMessages:` / `injectContext:` per team review
- **`automatic` + `searchTool` booleans** — replaces `mode: 'tool'` per team review
- **Plugin interface (SDK >=0.7.0)** — uses `initAgent()` + `getTools()`, one-line registration via `plugins: [memoryPlugin]`
- **Idempotent retries** — `clientToken` is generated once at buffer time and reused across the initial send and the retry, guaranteeing the service deduplicates correctly
- **Tool-use / reasoning blocks excluded from LTM** — only `textBlock` content reaches the memory store; agent internals don't pollute. Customers who need them can override via `messageFilter`.

### API shape

```typescript
import { AgentCoreMemory } from 'bedrock-agentcore/experimental/memory/strands'

const memoryPlugin = new AgentCoreMemory({
  memoryId: 'mem-1',
  actorId: 'user-1',
  sessionId: 'session-1',
  extraction: true,  // or { batchSize: 20, messageFilter: (msg) => ... }
  injection: {
    namespaces: {
      '/users/{actorId}/facts': { topK: 5 },
      '/users/{actorId}/preferences': { topK: 3 },
    },
    searchTool: true,
  },
})

const agent = new Agent({
  model,
  plugins: [memoryPlugin],
})
```

## Strands v2 compatibility

This plugin is the v1 user-facing surface. When Strands v2 ships its `MemoryManager` primitive, an `AgentCoreKnowledgeStore` (implementing the `KnowledgeStore` interface) will be added alongside this plugin — both surfaces will be supported, no breaking changes to existing users. The internal `AsyncBatcher` utility was designed to be promotable to a public export if a second consumer (telemetry, identity events, etc.) emerges.

## Related Issues

Design doc: https://quip-amazon.com/odNFAWZ9XaS9
Bug-bash test plan: https://quip-amazon.com/liRSADYLxnPv

## Type of Change

New feature

### Unit tests (541 / 541 pass)
- `npm run check` passes (lint, format, type-check, build, tests)
- Plugin class: 36 tests covering hook registration, injection pipeline, extraction pipeline, error handling, multi-agent helpers
- Format utilities: 44 tests covering all edge cases (regex escaping, SystemContentBlock[] handling, truncation, etc.)
- Search tool: 9 tests covering namespace search, filtering, error resilience
- **`AsyncBatcher`**: 15 tests covering size/time triggers, retry idempotency, timeout, drain dedup, maxDrainIterations cap, shutdown, callback error isolation
- **Regression suite (`_verify_concerns.test.ts`)**: 17 tests proving each previously-addressed PR review concern is still fixed (clientToken stability, template resolution, validation throws, lifecycle APIs, content-block filtering)

### E2E testing (deployed agent)
Deployed a real TypeScript agent to AgentCore with the plugin and verified:
- **Extraction** — messages stored as events, processed into LTM facts by the service
- **Injection** — cross-session recall of user facts (name, job, preferences) via system prompt injection
- **search_memory tool** — agent autonomously searched memory to find "TypeScript over Python" preference
- **STM (SessionManager + S3)** — conversation history persisted to S3 via Strands SessionManager, restored on next turn
- **Cross-session LTM** — brand new session recalled facts from prior sessions

Test project: `~/workplace/cli-typescript/memorytest/`

- [x] I ran `npm run check`

## Review feedback addressed

This PR has been through three rounds of review. Status of each comment:

**jariy17's evaluation (43-test edge-case + performance audit):**
- ✅ Finding 1 — namespace `{actorId}` / `{sessionId}` template resolution (now resolved at construct time)
- ✅ Finding 2 — tool-use / reasoning blocks pollute LTM (now excluded from `extractText`)
- ✅ Finding 3 — empty messages cause wasted `createEvent` calls (now skipped at buffer time via `shouldBuffer`)
- ✅ Finding 4 — `flush()` / `shutdown()` lifecycle APIs missing (now public methods)
- ✅ Performance note — `batchSize` semantics clarified in TSDoc on `ExtractionConfig`

**jesseturner21's inline review:**
- ✅ `clientToken` regenerated on retry (now generated once at buffer time and stored alongside the message)
- ✅ `memoryClient` type should be `MemoryClientConfig | MemoryClient` (done; `cloneWithConfig` no longer uses `Object.defineProperty`)

**Hweinstock's inline review:**
- ✅ Decouple batching from session manager — `AsyncBatcher<T>` extracted into `batcher.ts` (internal), 15 unit tests
- ✅ Explicit return type for `createSearchMemoryTool` — now `SearchMemoryTool = InvokableTool<SearchMemoryInput, string>`

## Checklist
- [ ] I have updated the documentation accordingly

**Note:** `@strands-agents/sdk` peer dependency bumped from `>=0.1.0` to `>=0.7.0` for the Plugin interface. Existing browser/code-interpreter integrations use `tool()` which exists in both versions, so this is backward compatible for those subpaths.

----

### DIFF
--- a/.workspace/notes/implementation-plan.md
+++ b/.workspace/notes/implementation-plan.md
@@ -1,0 +1,497 @@
+# AgentCoreMemory Plugin — Implementation Plan
+
+**Status:** Approved (Ralplan consensus: Planner → Architect → Critic)
+**Date:** 2026-04-22
+**Branch:** `feat/agentcore-memory` → feature branches per step
+
+---
+
+## RALPLAN-DR Summary
+
+### Principles
+1. Memory is never critical path — all failures log + continue, never crash the agent
+2. Follow existing SDK patterns — mirror integrations/strands/ directory structure
+3. Plugin interface is the contract — implement Strands Plugin (0.7.0)
+4. Extraction and injection are independent — customer opts into each separately
+5. Batch by default — reduce API call volume without customer effort
+
+### Decision Drivers
+1. API shape matches team consensus — `extraction:`/`injection:` naming, `automatic`/`searchTool` booleans
+2. MemoryClient is the only integration surface — `createEvent()` and `retrieveMemoryRecords()`
+3. Existing tool patterns set the structure — `src/memory/integrations/strands/`
+
+### Chosen Option: Single Plugin class + tool factory
+Single `AgentCoreMemory` class implementing Plugin interface, with internal helper functions (not classes) for formatting and batching. Also exports `createSearchMemoryTool` as a standalone factory for advanced composability.
+
+---
+
+## Naming Changes from Design Doc
+
+The following names were changed per team review consensus (Quip comments). The design doc examples still use the old names and will need updating:
+
+| Design Doc | Implementation | Rationale |
+|---|---|---|
+| `AgentCoreMemoryPlugin` | `AgentCoreMemory` | Jack Yuan: "Plugin doesn't need Plugin in the name" |
+| `storeMessages:` | `extraction:` | Team consensus: options 1 or 3 from Quip thread |
+| `injectContext:` | `injection:` | Parallel naming with extraction |
+| `mode: 'tool'` | `automatic: false, searchTool: true` | Jesse's `automaticInjection` suggestion (2 likes) |
+| `bedrock-agentcore/memory/strands` | `bedrock-agentcore/experimental/memory/strands` | Matches existing SDK export convention |
+
+## Design Decision Resolutions
+
+Open questions from the design doc, resolved during review:
+
+| Question | Resolution | Source |
+|---|---|---|
+| Prefetch mode needed on day one? | No — deferred until customer demand | Design doc Decision 1 recommendation, no objection |
+| searchTool auto vs explicit opt-in? | Explicit opt-in (`searchTool: false` default) | Design doc recommendation adopted |
+| batchSize default? | 10 (match Python SDK) | Design doc Decision 2, no objection |
+| Built-in filter presets? | Deferred — leave filtering to customer via `messageFilter` | No team feedback requesting presets |
+| Auto-derive namespace labels? | Yes, from path | Design doc Decision 3 recommendation adopted |
+| Staleness metadata in injection? | Deferred — not enough signal to justify extra tokens | No team feedback requesting it |
+| contextTag configurability? | Yes — TJ requested it | Quip comment, moved from open question |
+| System prompt reinforcement for tool mode? | No — TJ + Mackenzie consensus | Quip comment: "shouldn't interfere with customer's system prompt" |
+| Await vs fire-and-forget flush? | Await by default, `fireAndForget: true` escape hatch | Aidan's decision |
+
+## Why BeforeInvocationEvent, Not BeforeModelCallEvent
+
+We considered `BeforeModelCallEvent` for injection — it fires after user message is appended, so the current user message would be available as a search query. We rejected it because `BeforeModelCallEvent` fires on **every model call within an invocation**, including after each tool-use loop iteration. With ephemeral injection, this would mean stripping and re-injecting memory context (with a retrieval API call) on every loop iteration — wasteful and potentially 5-10x the API calls per turn for tool-heavy agents. `BeforeInvocationEvent` fires once per invocation, which is the right granularity.
+
+## Verified Strands SDK 0.7.0 Facts
+
+- `Plugin` interface: `{ readonly name: string; initAgent(agent: LocalAgent): void | Promise<void>; getTools?(): Tool[] }`
+- `AgentConfig` has `plugins?: Plugin[]`
+- `initAgent()` receives the `Agent` instance at runtime (typed as `LocalAgent`, but Agent passes `this`)
+- `systemPrompt` is on `Agent`, NOT on `LocalAgent` — must cast in initAgent
+- `BeforeInvocationEvent` fires BEFORE user message is appended to `agent.messages`
+- **`MessageAddedEvent` DOES fire for user input** despite JSDoc claiming otherwise. Verified: `_appendMessage()` is called for user messages and yields `MessageAddedEvent`. The JSDoc is stale.
+- `AfterInvocationEvent` fires in the `finally` block (always runs)
+- `SystemPrompt = string | SystemContentBlock[]` where `SystemContentBlock = TextBlock | CachePointBlock | GuardContentBlock`
+
+## MemoryClient API Surface (PR #139, merged)
+
+- `createEvent({ memoryId, actorId, sessionId, eventTimestamp, payload, metadata?, clientToken? })` — single event, NO batch event API. `payload` is `PayloadType[]` (array — can hold multiple conversational items per event). `clientToken` enables idempotent retries.
+- `retrieveMemoryRecords({ memoryId, namespace, searchCriteria: { searchQuery, topK? } })` — `namespace` is a **prefix filter** (not exact match). One call per namespace needed for per-namespace `topK` control. Requires `searchQuery` string.
+- Payload: `PayloadType[] where ConversationalMember = { conversational: { role: Role, content: { text: string } } }`
+- Role: AWS SDK uses uppercase `'USER' | 'ASSISTANT' | 'TOOL' | 'OTHER'`. Strands uses lowercase `'user' | 'assistant'`.
+- MetadataValue: discriminated union `{ stringValue: string; $unknown?: never }` from `@aws-sdk/client-bedrock-agentcore`
+
+---
+
+## Step 1: Types & Configuration
+
+**File:** `src/memory/integrations/strands/types.ts`
+**Tests:** Type-only — TypeScript compiler validates
+
+```typescript
+export interface NamespaceConfig {
+  topK?: number              // default: 5
+  relevanceScore?: number    // client-side post-filter on response scores (NOT an API param)
+                             // records below this threshold are dropped after retrieval
+}
+
+export interface ExtractionConfig {
+  batchSize?: number              // default: 10
+  batchTimeoutMs?: number         // default: 5000
+  messageFilter?: (message: Message) => boolean
+  fireAndForget?: boolean         // default: false (await flush)
+}
+
+export interface InjectionConfig {
+  namespaces: Record<string, NamespaceConfig>
+  automatic?: boolean              // default: true
+  searchTool?: boolean             // default: false
+  maxInjectionChars?: number       // default: 8000 (~2000 tokens)
+  contextTag?: string              // default: 'agentcore_memory'
+  formatMemories?: (records: MemoryRecordGroup[]) => string
+}
+
+export interface AgentCoreMemoryConfig {
+  memoryId: string
+  actorId: string
+  sessionId: string
+  extraction?: boolean | ExtractionConfig    // Resolution: undefined/false = disabled,
+                                             // true/{} = enabled with defaults,
+                                             // { batchSize: 20 } = enabled with overrides
+  injection?: InjectionConfig
+  metadataProvider?: (message: Message) => Record<string, MetadataValue>
+  memoryClient?: MemoryClientConfig
+}
+
+export interface MemoryRecordGroup {
+  namespace: string
+  label: string              // auto-derived: '/facts/{actorId}/' → 'facts'
+  records: MemoryRecord[]
+}
+
+export interface MemoryRecord {
+  content: string
+  score?: number
+  createdAt?: Date
+}
+
+// Internal — defaults applied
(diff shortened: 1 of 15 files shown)


.workspace/notes/implementation-plan.md
```diff
@@ -0,0 +1,497 @@
+# AgentCoreMemory Plugin — Implementation Plan
```
jesseturner21, line 1: side not i actually like having the implementation plan in the repo. And i think some way this will be added into the regular dev flow

src/memory/integrations/strands/plugin.ts
```diff
@@ -0,0 +1,293 @@
...
+    this.buffer = []
+    this.clearFlushTimer()
+
+    try {
+      const results = await Promise.allSettled(toFlush.map((msg, i) => this.createEventFromMessage(msg, i)))
+
+      const failed = results.map((r, i) => (r.status === 'rejected' ? i : null)).filter((i): i is number => i !== null)
+
+      if (failed.length > 0) {
+        const retryResults = await Promise.allSettled(failed.map((i) => this.createEventFromMessage(toFlush[i]!, i)))
```
jesseturner21: client token is regenerated here for every create event from message. If the create fails then the retry will create a new one with a new client token. This defeats the purpose of idempotency. Then a user could have two of the same event in memory. To fix this have create from message accept a client token and then pass that in instead of generating it in the create 

eg. 
``` 
private async flushBuffer(): Promise<void> {
    if (this.flushing) return
    if (this.buffer.length === 0) return

    this.flushing = true
    const toFlush = [...this.buffer]
    this.buffer = []
    this.clearFlushTimer()

    const tokens = toFlush.map((_, i) => this.generateClientToken(i))

    try {
      const results = await Promise.allSettled(
        toFlush.map((msg, i) => this.createEventFromMessage(msg, tokens[i]!))
      )

      const failed = results
        .map((r, i) => (r.status === 'rejected' ? i : null))
        .filter((i): i is number => i !== null)

      if (failed.length > 0) {
        const retryResults = await Promise.allSettled(
          failed.map((i) => this.createEventFromMessage(toFlush[i]!, tokens[i]!))
        )
        // ...
      }
    } finally {
      this.flushing = false
    }
  }

  And update createEventFromMessage to accept the token as a parameter instead of generating it internally:

  private createEventFromMessage(message: Message, clientToken: string): Promise<unknown> {
    // ... use clientToken directly instead of calling this.generateClientToken()
  }

```

src/memory/integrations/strands/types.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export interface AgentCoreMemoryConfig {
+  memoryId: string
+  actorId: string
+  sessionId: string
+  extraction?: boolean | ExtractionConfig
+  injection?: InjectionConfig
+  metadataProvider?: (message: Message) => Record<string, { stringValue: string }>
+  memoryClient?: MemoryClientConfig
```
jesseturner21: This type should be MemoryClientConfig | MemoryClient — the constructor already handles both via an instanceof check
  [(ref)](https://github.com/aws/bedrock-agentcore-sdk-typescript/blob/3f6da2fdff576ebbc3f446b2875f2db9392d3478/src/memory/integrations/strands/plugin.ts#L50).

  Accepting both types would clean up cloneWithConfig. Currently it creates a throwaway client:

  ```typescript
  // Current: creates a MemoryClient in constructor, then immediately overwrites it
  const instance = new AgentCoreMemory(config)
  Object.defineProperty(instance, 'client', { value: source.client })
```


  If the type accepted MemoryClient, it could just pass it through:

```typescript
  // Cleaner: no wasted allocation, no Object.defineProperty hack
  const instance = new AgentCoreMemory({ ...config, memoryClient: source.client })

**Review by jesseturner21 — APPROVED** [review]

**Comment by jariy17** [comment]
## AgentCoreMemory Plugin — Edge Case & Performance Evaluation (PENDING)

We ran a comprehensive evaluation of this plugin with 43 custom tests (performance benchmarks + edge case probes) on top of the existing 100 unit tests. All existing tests pass. Below are the findings that are **probable in real-world production usage**.

---

### Finding 1: Namespace template placeholders (`{actorId}`, `{sessionId}`) are not resolved

**Severity: High — silently breaks retrieval**

The design doc and PR description show namespace paths like `/users/{actorId}/facts`. The plugin passes these as **literal strings** to `retrieveMemoryRecords` — no string replacement occurs anywhere in the codebase.

```typescript
// plugin.ts:131 — namespace sent as-is
const response = await this.client.retrieveMemoryRecords({
  memoryId: this.originalConfig.memoryId,
  namespace: ns,  // ← "/users/{actorId}/facts" — never resolved
  searchCriteria: { ... },
})
```

**Impact**: Injection and `search_memory` query a nonexistent namespace → return empty results → agent has no memory. Silent failure, no error logged.

This also means `withActor('bob')` doesn't actually change the queried namespace — both the original and the fork query the same literal `{actorId}` path.

**Suggested fix**: Resolve templates at construct time (and in `withActor`/`cloneWithConfig`):

```typescript
function resolveNamespaces(
  namespaces: Record<string, NamespaceConfig>,
  actorId: string,
  sessionId: string,
): Record<string, NamespaceConfig> {
  const resolved: Record<string, NamespaceConfig> = {}
  for (const [ns, config] of Object.entries(namespaces)) {
    resolved[ns.replace('{actorId}', actorId).replace('{sessionId}', sessionId)] = config
  }
  return resolved
}
```

---

### Finding 2: Tool use and reasoning blocks are serialized into extraction events

**Severity: Medium — pollutes long-term memory store**

`extractText()` in `format.ts` serializes ALL content block types into the `createEvent` payload:

```typescript
// format.ts:110-111
case 'toolUseBlock':
  return `[tool_use: ${block.name}(${JSON.stringify(block.input)})]`
case 'reasoningBlock':
  return block.text ?? undefined
```

For a typical tool-using agent turn, this means LTM events contain:
- `"Let me think about this carefully..."` (internal chain-of-thought)
- `"[tool_use: search({"query":"weather in SF"})]"` (internal mechanics)
- Raw tool result JSON

These are agent internals, not user-relevant memories. They pollute the memory store and add noise to future retrievals.

**Suggested fix**: Either filter out `toolUseBlock`/`toolResultBlock`/`reasoningBlock` types in `extractText()`, or add a configurable `contentFilter` option to extraction config so customers can control what gets stored.

---

### Finding 3: Empty/non-text messages generate wasted `createEvent` API calls

**Severity: Medium — unnecessary cost**

Messages with no text content (empty content array, image-only, or unknown block types) still trigger `createEvent` with `content: { text: "" }`.

Verified empirically:
```
[EDGE-EMPTY] Payload text: {"text":""}    ← empty content array
[EDGE-IMAGE] Sent text: ""                ← image-only message
```

For tool-heavy agents, this means several wasted API calls per turn with no useful data stored.

**Suggested fix**: Skip `createEvent` when `extractText()` returns an empty string:

```typescript
private createEventFromMessage(message: Message, index: number): Promise<unknown> {
  const text = extractText(message)
  if (!text) return Promise.resolve() // ← skip empty messages
  // ...
}
```

---

### Finding 4: `flush()` and `shutdown()` lifecycle APIs are missing

**Severity: Medium — data loss risk with `fireAndForget: true`**

The design doc documents these as public instance methods:
```
plugin.flush(): Promise<void>     // force immediate drain of buffered messages
plugin.shutdown(): Promise<void>  // flush + mark inert, stop accepting messages
```

Neither exists on the class. With `fireAndForget: true`, the `AfterInvocationEvent` flush runs in the background and is not awaited — if the container shuts down before it completes, buffered messages are lost. Without `shutdown()`, there's no way to hook into `SIGTERM` for graceful drain.

With `fireAndForget: false` (default), `AfterInvocationEvent` awaits the flush, so this is not an issue for default config.

---

### Performance: AWS API Call Profile

We instrumented the mock client to count exact API calls. Results:

| Scenario | `retrieveMemoryRecords` | `createEvent` | Total |
|---|---|---|---|
| Single turn, 1 namespace | 1 | 2 | **3** |
| Single turn, 3 namespaces | 3 | 2 | **5** |
| 10-turn conversation, 2 namespaces | 20 | 20 | **40** |
| Worst case with retries (all fail once) | 20 | 40 | **60** |

Note: `batchSize` controls **when** the flush triggers, not API call count. Each message is always 1 `createEvent` call — there is no true batching. This is worth documenting so customers understand the cost profile.

---

*Evaluation ran 43 custom tests (13 performance + 30 edge case) — all passing. Test files available on request.*

jariy17:
Test files for the evaluation above:

https://gist.github.com/jariy17/e427dd126eac9872b901b04afa18c6dd

- `performance.test.ts` — 13 tests: API call counts per operation, retry overhead, batchSize impact, fireAndForget behavior
- `edge-cases.test.ts` — 30 tests: buffer concurrency, clientToken uniqueness, empty/large messages, metadataProvider failures, namespace templates, input validation, format edge cases, missing lifecycle APIs

To run: `npx vitest run src/memory/integrations/strands/__tests__/performance.test.ts src/memory/integrations/strands/__tests__/edge-cases.test.ts`

src/memory/integrations/strands/plugin.ts
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    if (!this.flushTimer && this.buffer.length === 1) {
+      this.flushTimer = globalThis.setTimeout(() => {
+        this.flushBuffer().catch((err) => console.warn('[agentcore-memory] Timer flush failed:', err))
+      }, this.extractionConfig!.batchTimeoutMs)
+    }
+  }
+
+  private async flushBuffer(): Promise<void> {
```
Hweinstock: late to the PR, but wondering if it makes sense to decouple the batching logic from the session manager. I feel like batching utilities could become a general and re-usable construct. (could be done in the future)

src/memory/integrations/strands/search-memory-tool.ts
```diff
@@ -0,0 +1,73 @@
...
+  const parts = ns.split('/').filter((p) => p.length > 0 && !p.startsWith('{'))
+  return parts[parts.length - 1] ?? ns
+}
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type, @typescript-eslint/explicit-module-boundary-types
+export function createSearchMemoryTool(
+  client: MemoryClient,
+  config: { memoryId: string; namespaces: Record<string, NamespaceConfig> }
+) {
+  return tool({
```
Hweinstock, line 27: Is there a strong reason for why we don't want a return type here?

PR #176, opened 2026-05-19
test: add OTEL span content leakage integration tests

jesseturner21:
## Summary

- Adds integration tests that assert user content (prompts, agent responses) never appears in OTEL spans
- COE Finding: prevents user content from leaking into observability backends via span attributes/events
- Tests use sentinel strings sent through the full request lifecycle, then verify absence from all exported span data
- Currently passes trivially (SDK does not emit OTEL spans yet) — acts as regression guard for when OTEL is added

## Test cases

- `sync invocation does not leak user content into spans` — standard JSON request/response
- `streaming invocation does not leak user content into spans` — async generator streaming
- `websocket invocation does not leak user content into spans` — WebSocket bidirectional messaging

## Changes

- Added `@opentelemetry/api`, `@opentelemetry/sdk-trace-base`, `@opentelemetry/sdk-trace-node` to devDependencies
- New test file: `tests_integ/otel-no-user-content.test.ts`

## Test plan

- [x] All 3 tests pass locally
- [ ] CI passes


tests_integ/otel-no-user-content.test.ts
```diff
@@ -0,0 +1,198 @@
...
+    let app: BedrockAgentCoreApp
+    let fastify: any
+
+    beforeEach(async () => {
+      const handler: InvocationHandler = async (req, _context) => {
+        return { response: SENTINEL_RESPONSE, echo: (req as any).prompt }
+      }
+
+      app = new BedrockAgentCoreApp({ invocationHandler: { process: handler } })
+      fastify = (app as any)._app
```
Hweinstock, line 81: nit: it would be great to avoid this casting, but looking at the implementation I think its OOS to refactor.

Hweinstock (approved):

tests_integ/otel-no-user-content.test.ts
jesseturner21: true exposing the fastify app is OOS and would need a discussion if that is something we want to expose.

PR #187, opened 2026-06-10
feat(memory): add AgentCore Memory Strands integration (experimental)

aidandaly24:
## Description

Adds an **experimental** AgentCore Memory integration for the Strands TypeScript SDK: `AgentCoreMemoryStore`, a `MemoryStore` that plugs into Strands' `MemoryManager`. Exposed via the package export `bedrock-agentcore/experimental/memory/strands`.

- **Recall** — `search()` maps to `retrieveMemoryRecords`. The read target is a discriminated union: `{ namespace }` (exact prefix) or `{ namespacePath }` (subtree), mirroring the API directly. Optional client-side `minScore` floor with over-fetch, and reserved underscored result-metadata keys (`_id`/`_score`/`_namespaces`/`_createdAt`).
- **Write** — `addMessages()` packs a turn's role-tagged messages into a single `createEvent` (one API call carrying many turns, chunked at `maxTurnsPerEvent`); AgentCore extracts and consolidates server-side (no client-side LLM pass). Idempotent via a run-anchored `clientToken` over the message sequence range.
- **Construction** — the store stands alone with flat config: `new AgentCoreMemoryStore({ memoryId, actorId, sessionId, namespace, writable?, extraction? })` (self-names from the namespace; `writable` defaults to `false`; creates its own client if none is passed). For multiple namespaces, `createAgentCoreMemoryStores(...)` builds one shared client and returns stores ready to spread into `MemoryManagerConfig.stores`.
- **Write topology** — `createEvent` is namespace-free, so at most one store per `(actorId, sessionId)` may be writable. The writer is the namespace flagged `writable: true` (else the first when extraction is enabled); `assertWritableTopology` enforces the at-most-one rule and is exported so hand-built multi-store setups can self-guard.
- **Cadence** — `extraction` is the single write switch (`boolean | { cadence?, filter? }`): omit for recall-only, `true` for the MemoryManager's default cadence, or pass any Strands `ExtractionTrigger` (single or array) for custom cadence. Write-cost reduction comes from payload batching (always on); cadence and `MemoryManager.flush()` control when/how durably writes land.

Bumps `@aws-sdk/client-bedrock-agentcore` to `^3.1065.0` for the typed `namespacePath` field, and depends on `@strands-agents/sdk` `>=1.5.0` for the `MemoryManager` / `MemoryStore` / extraction interface.

> **Status: experimental.** The module is GA-blocked only by the usual experimental-surface bar, not by any upstream dependency — the Strands memory interface is published (`@strands-agents/sdk` 1.6.0) and imported directly (no vendored mirror).

## Related Issues

Implements the Strands memory extraction interface from `strands-agents/harness-sdk` #2671 (now published in `@strands-agents/sdk`).

## Review history (this PR)

The implementation went through multiple review rounds and was hardened before this description was rewritten:

- **Factory→store refactor** (review feedback): the store mirrors `retrieveMemoryRecords` via the `{ namespace } | { namespacePath }` union (dropped the `readMode` flag), self-names, and stands alone with flat config; the singular `createAgentCoreMemoryStore` was removed in favor of the constructor; `assertWritableTopology` extracted + exported; the AgentCore-specific batch trigger was removed (generic cadence belongs upstream — the `cadence` field accepts any Strands trigger).
- **Code review + multi-agent bug bash** of the whole module surfaced and fixed: a call-time `maxSearchResults` validation gap (a `0` from the `search_memory` tool silently returned `[]`); metadata values of `undefined`/`NaN`/`Infinity`/`null` slipping past the charset guard; `$`-sequence namespace corruption in `resolveNamespace`; an unmatched-brace gap in namespace validation; an opaque `TypeError` on non-string identity fields; a silently-overridden explicit `writable: false`; and a spurious `extraction: false` warning. All fixed with tests; core read/write/recall logic held across the bash.

## Documentation

Included in this PR: `docs/MEMORY.md` (concept + deployment guide) and `src/memory/integrations/strands/README.md` (API reference). Main `README.md` Features entry lists Memory as experimental → `docs/MEMORY.md`.

## Type of Change

New feature (experimental)

## Testing

- [x] **111 unit tests** in `src/memory/integrations/strands/__tests__/` (store search/topK/minScore + read-target mapping; sender batching/chunking/clientToken/metadata-charset/grouping/AggregateError; factory writer-selection + `assertWritableTopology` + validation; namespace resolution + `$`-safety + brace/placeholder rejection; format helpers). Full unit suite green.
- [x] **9 live integration tests** in `tests_integ/memory.test.ts` against real AgentCore Memory (deploy account): write → server-side extraction → recall round-trip, idempotency, multi-turn batching = one `createEvent`, the namespace contract (records land under the queried `{actorId}` path), `{sessionId}` per-session isolation, the standalone directly-constructed store, and a full `MemoryManager` + `Agent` E2E. All passing in CI (the memory suite polls live extraction, which is slow — the integ step timeout was raised to 25m to accommodate it).
- [x] `tsc --noEmit`, ESLint, and Prettier clean on the module.

## Checklist
- [x] I have added an appropriate example to the documentation to outline the feature
- [x] Any dependent changes have been merged and published (`@strands-agents/sdk` memory interface is published and imported directly)

----


docs/MEMORY.md
```diff
@@ -0,0 +1,146 @@
...
+- **Consolidation is the dedup backstop.** Because consolidation merges facts at the record level, a
+  duplicate *event* (e.g. from a retry) does not become a duplicate long-term memory.
+
+## Quick start
+
+```typescript
+import { Agent, MemoryManager, MessageAddedEvent } from '@strands-agents/sdk'
+import { createAgentCoreMemoryStores, AgentCoreBatchTrigger } from 'bedrock-agentcore/experimental/memory/strands'
+
+const stores = createAgentCoreMemoryStores({
```
opieter-aws, line 47: Could we create a thin convenience wrapper in the case a user wants to set up a single store?

``` typscript
const store = createAgentCoreMemoryStore({          // singular, returns one store
  memoryId, actorId, sessionId,
  namespace: '/users/{actorId}/facts',
})
```

reads much more intuitive if I have just 1 store, which I assume is the majority of the use cases here.

src/memory/strands/store.ts
```diff
@@ -0,0 +1,146 @@
...
+    })
+
+    let summaries: MemoryRecordSummary[]
+    try {
+      const out = await this.client.send(command)
+      summaries = out.memoryRecordSummaries ?? []
+    } catch (err) {
+      // Fail-open: a failed recall must never break the agent loop. Return nothing for this store.
+      console.warn(`[agentcore-memory] search failed for store "${this.name}":`, err)
+      return []
```
opieter-aws: The memory manager ensure the agent loop doesn't break here, so you can throw. Otherwise failures are swallowed on the MM level

src/memory/strands/sender.ts
```diff
@@ -0,0 +1,136 @@
...
+      if (timer) globalThis.clearTimeout(timer)
+    }
+  }
+
+  private reportDropped(message: MessageData, reason: DropReason, cause?: unknown): void {
+    try {
+      this.onDropped?.({ reason, text: extractText(message), ...(cause !== undefined && { cause }) })
+    } catch (err) {
+      // Never let a customer callback break the write path.
+      console.warn('[agentcore-memory] onDropped callback threw:', err)
```
opieter-aws: Can you use the SDK logger here instead?

src/memory/strands/types.ts
```diff
@@ -0,0 +1,111 @@
...
+  /** Pre-constructed AWS client (e.g. shared across a set of stores). */
+  readonly client?: BedrockAgentCoreClient | undefined
+}
+
+/**
+ * Config for a single {@link AgentCoreMemoryStore}: the shared {@link AgentCoreMemoryConfig} plus this
+ * store's per-namespace identity. The factory builds the shared config once and one of these per
+ * namespace.
+ */
+export interface AgentCoreMemoryStoreConfig {
```
opieter-aws: Can this extend MemoryStoreConfig?

docs/MEMORY.md
```diff
@@ -0,0 +1,146 @@
...
+  memoryId: process.env.MEMORY_MYMEMORY_ID!, // injected by the deploy (see below)
+  actorId, // see "Who is the actor?"
+  sessionId, // from the runtime request context
+  namespaces: [
+    { namespace: '/users/{actorId}/facts' },
+    { namespace: '/users/{actorId}/preferences' },
+  ],
+  trigger: new AgentCoreBatchTrigger({ messageCount: 4, maxDelayMs: 3000, messageAddedEvent: MessageAddedEvent }),
```
opieter-aws: Can we to converge the config surface here? Currently there is some tension in that the `trigger` is flat (trigger for what?), it's mandatory (no defaults), and there are 3 things that can disagree: `writable`, the flat mandatory trigger, and `writeNamespace`.

For AC, `writable` ⟺ extraction-enabled ⟺ “writes via createEvent”. So can we collapse these into one switch, reusing the upstream `boolean | config` setup?

```typescript
extraction?: boolean | { cadence?: ExtractionTrigger; namespace?: string }
// omitted/false -> recall-only   
// true -> writable, default cadence
//
// { cadence }   -> custom cadence  
// { namespace } -> which namespace writes

// today
{ writeNamespace: '...', trigger: new AgentCoreBatchTrigger({ messageAddedEvent: MessageAddedEvent }) }
// proposed shorthand convenience
{ extraction: true }
```

This would give it one mental model and make recall-only expressible

WDYT?

docs/MEMORY.md
```diff
@@ -0,0 +1,146 @@
...
+- **Resource setup is out of scope.** Strategies, indexed keys, record streaming, expiry, and
+  encryption are control-plane concerns configured at `CreateMemory` (via the CLI/CDK), not through
+  this store.
+
+## Known limitations
+
+1. **Metadata-filtered recall (indexed keys) is not exposed to the agent.** AgentCore supports
+   `metadataFilters` on retrieval (gated on indexed keys declared at resource creation), but the
+   generic `MemoryStore.search(query, { maxSearchResults })` interface has no slot to carry a filter,
+   so the `search_memory` tool can't express one. Pending an upstream `SearchOptions` addition.
```
opieter-aws: nit: per our discussion, this needs to be set as per-instance defaults on the store

docs/MEMORY.md
aidandaly24: Added `createAgentCoreMemoryStore` (singular) returning one store for the single-namespace case — 519d6c9. Same options, just `namespace` instead of `namespaces`.

src/memory/strands/store.ts
aidandaly24: Good catch — confirmed `MemoryManager.search` wraps each store in `Promise.allSettled`, so I removed the try/catch and let `search()` throw. Errors are now isolated per-store and surfaced by the manager instead of swallowed — 519d6c9.

src/memory/strands/sender.ts
aidandaly24: Done — routed through a module `Logger` (with `configureMemoryLogging` to inject) instead of `console`, mirroring the Strands logger convention — 519d6c9. (`@strands-agents/sdk` exports `configureLogging`/`Logger` but not the `logger` instance, so an external package owns a console-backed default of the same shape.)

src/memory/strands/types.ts
aidandaly24: Done — `AgentCoreMemoryStoreConfig extends MemoryStoreConfig` now (matching `BedrockKnowledgeBaseStoreConfig`), inheriting name/description/maxSearchResults/extraction — 519d6c9.

docs/MEMORY.md
aidandaly24: Agreed — converged onto a single `extraction?: boolean | { cadence?, namespace? }` switch, removing the separate `trigger`/`writeNamespace` — 519d6c9. Omitting it now gives recall-only (which wasn't expressible before); `true` uses the default cadence; the object form sets custom cadence / write namespace. One pre-publish wrinkle: `true` needs a `messageAddedEvent` to build the default trigger (since the real `MessageAddedEvent` isn't importable until the flip), so it's a top-level factory input for now.

docs/MEMORY.md
aidandaly24: Fixed — reworded to 'per-instance store defaults' (bake the filter into store config like `minScore`), and dropped the stale 'pending SearchOptions' line since that addition was dropped upstream — 519d6c9.

src/memory/strands/factory.ts
```diff
@@ -0,0 +1,282 @@
...
+  }
+
+  const readMode: ReadMode = input.readMode ?? 'per-namespace'
+
+  // Resolve the single `extraction` switch into: is writing enabled, with what cadence, on which namespace.
+  const writeEnabled = input.extraction !== undefined && input.extraction !== false
+  const extractionObj: AgentCoreExtractionConfig = typeof input.extraction === 'object' ? input.extraction : {}
+  const extraction: ExtractionConfig | undefined = writeEnabled
+    ? { trigger: extractionObj.cadence ?? defaultTrigger(input.messageAddedEvent) }
+    : undefined
```
opieter-aws: For `extraction: true`, can we pass the boolean straight through to the store’s extraction field instead of eagerly building a trigger here? The MemoryManager defaults the trigger to the SDK’s own `IntervalTrigger` using its internal event import, so the default path needs no `messageAddedEvent`. Then you can keep `AgentCoreBatchTrigger` only for the explicit `{ cadence }` case . Trade-off: the default becomes turn-based rather than byte/time (which seems fine for “just give me defaults", since that then syncs with Strands SDK). WDYT?

src/memory/strands/sender.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    const results = await Promise.allSettled(sendable.map((m) => this.sendOne(m)))
+
+    const failed = sendable.filter((_, i) => results[i]!.status === 'rejected')
+    if (failed.length === 0) return
+
+    const retried = await Promise.allSettled(failed.map((m) => this.sendOne(m)))
+    retried.forEach((r, i) => {
+      if (r.status === 'rejected') {
+        this.reportDropped(failed[i]!.message, 'retry-failed', (r as PromiseRejectedResult).reason)
```
opieter-aws: Failures are swallowed here. Because `addMessages` never throws, the `ExtractionCoordinator` never rolls back or re-fires. Consider throwing to make use of the coordinator’s retry

src/memory/strands/sender.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  /**
+   * Deterministic when a stable `seq` is present (re-fire-safe idempotency); otherwise undefined
+   * (v1: tolerate error-path duplicates, consolidation is the backstop). Never time/random-based,
+   * which would defeat dedup across a re-fire.
+   */
+  private clientTokenFor(item: SeqMessage): string | undefined {
+    if (item.seq === undefined) return undefined
+    return `${this.memoryId}-${this.actorId}-${this.sessionId}-${item.seq}`
```
opieter-aws: I confirmed that sessionId does _not_ refresh upon session restore. Since the memory manager currently does not support session history, the `seq` counter is set back zero on a restored session. That makes this identifier not unique to the message. What's the `clientToken` dedup window? 

I would like to introduce unique message IDs in Strands which would solve this problem (we'd replace the `seqNumber` with a `uuid`), but we do not have this available today.

opieter-aws: As an interim mitigation, you could generate a run-unique ID at sender construction time and replace the `sessionId` with that

src/memory/strands/_strands-memory-types.ts
opieter-aws, line 1: The code has been released, so you can install the [latest version](https://github.com/strands-agents/harness-sdk/releases/tag/typescript%2Fv1.5.0)

src/memory/strands/store.ts
```diff
@@ -0,0 +1,140 @@
...
+      .filter((r) => this.minScore == null || (r.score ?? 0) >= this.minScore)
+      .slice(0, want)
+      .map((r) => this.toEntry(r))
+  }
+
+  private toEntry(record: MemoryRecordSummary): MemoryEntry {
+    const metadata: Record<string, JSONValue> = {}
+    if (record.memoryRecordId !== undefined) metadata.id = record.memoryRecordId
+    if (record.score !== undefined) metadata.score = record.score
+    if (record.namespaces !== undefined) metadata.namespaces = record.namespaces
```
opieter-aws: Consider using underscored keys for these reserved metadata fields to avoid potential collision with user-defined metadata

src/memory/strands/batch-trigger.ts
```diff
@@ -0,0 +1,111 @@
...
+        timer = undefined
+      }
+    }
+
+    const fire = (): void => {
+      reset()
+      context.fire() // fire-and-forget; coordinator only processes messages past the high-water mark
+    }
+
+    context.agent.addHook(this.messageAddedEvent, (event: unknown) => {
```
opieter-aws: Consider counting only `isUserOrAssistantWithText` messages here to match the Strands `perTurn` count. That would make it match with the memory trigger for other stores.

src/memory/strands/types.ts
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * How a store's `search()` targets AgentCore long-term records.
+ * - `'per-namespace'`: query one exact namespace prefix (`namespace`); one store per namespace.
+ * - `'subtree'`: query a parent path hierarchically (`namespacePath`), covering all child namespaces.
+ */
+export type ReadMode = 'per-namespace' | 'subtree'
+
+/** Per-message metadata attached to each `createEvent`. Values use the AgentCore metadata shape. */
+export type MetadataProvider = (message: MessageData) => Record<string, { stringValue: string }>
```
opieter-aws: Is it an idea to make this more lenient (`Record<string, JSONValue>`) and map to string internally?

src/memory/strands/factory.ts
```diff
@@ -0,0 +1,282 @@
...
+  if (input.namespaces.length === 0) {
+    throw new Error('createAgentCoreMemoryStores: at least one namespace is required')
+  }
+
+  const readMode: ReadMode = input.readMode ?? 'per-namespace'
+
+  // Resolve the single `extraction` switch into: is writing enabled, with what cadence, on which namespace.
+  const writeEnabled = input.extraction !== undefined && input.extraction !== false
+  const extractionObj: AgentCoreExtractionConfig = typeof input.extraction === 'object' ? input.extraction : {}
+  const extraction: ExtractionConfig | undefined = writeEnabled
```
opieter-aws: Is the filter passed through correctly here?

src/memory/integrations/strands/store.ts
```diff
@@ -0,0 +1,140 @@
...
+    // throw is isolated to this store and surfaced through the manager's partial-failure handling
+    // rather than swallowed here.
+    const out = await this.client.send(command)
+    return (out.memoryRecordSummaries ?? [])
+      .filter((r) => this.minScore == null || (r.score ?? 0) >= this.minScore)
+      .slice(0, want)
+      .map((r) => this.toEntry(r))
+  }
+
+  private toEntry(record: MemoryRecordSummary): MemoryEntry {
```
opieter-aws, line 178: Consider setting `createdAt` timestamp on the entries

src/memory/integrations/strands/store.ts
```diff
@@ -0,0 +1,140 @@
...
+  private readonly minScore?: number
+  private readonly sender?: AgentCoreEventSender
+
+  constructor(storeConfig: AgentCoreMemoryStoreConfig) {
+    const { config } = storeConfig
+    this.name = storeConfig.name
+    if (storeConfig.description !== undefined) this.description = storeConfig.description
+    if (storeConfig.maxSearchResults !== undefined) this.maxSearchResults = storeConfig.maxSearchResults
+    this.writable = storeConfig.writable
+    if (storeConfig.writable && storeConfig.extraction !== undefined) this.extraction = storeConfig.extraction
```
opieter-aws: Consider logging a warning here, so extraction is not dropped silently if writable is `false`

src/memory/strands/factory.ts
```diff
@@ -0,0 +1,282 @@
...
+  for (let i = 0; i < first.length; i++) {
+    const seg = first[i]
+    if (split.every((parts) => parts[i] === seg)) prefix.push(seg!)
+    else break
+  }
+  const joined = prefix.join('/')
+  return joined.length > 0 ? joined : undefined
+}
+
+function ensureUniqueNames(stores: AgentCoreMemoryStore[]): void {
```
opieter-aws: You can defer this check to the memory manager: https://github.com/strands-agents/harness-sdk/blob/main/strands-ts/src/memory/memory-manager.ts#L110

jona62:
I think we should have this under `memory/integrations/strands` as opposed to the flat structure `memory/strands`

src/memory/strands/batch-trigger.ts
```diff
@@ -0,0 +1,111 @@
...
+      n++
+      if (maxBytes !== undefined && hasMessage(event)) {
+        bytes += extractText(event.message).length
```
jona62: `.length` here doesn't return bytes, you might want something like this `Buffer.byteLength(text, 'utf8')`

src/memory/strands/sender.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  private async withTimeout(send: Promise<unknown>, message: MessageData): Promise<void> {
+    let timer: ReturnType<typeof globalThis.setTimeout> | undefined
+    const timeout = new Promise<never>((_, reject) => {
+      timer = globalThis.setTimeout(() => {
```
jona62:
```
Bug — timeout reports a false / duplicate drop (sender.ts, withTimeout)

  withTimeout calls reportDropped inside the timer callback, before the retry runs:

  timer = globalThis.setTimeout(() => {
    this.reportDropped(message, 'timeout')   // fires per timed-out attempt, regardless of outcome
    reject(new Error(...))
  }, this.sendTimeoutMs)

  Two failure modes:
  1. False positive — attempt 1 times out but the retry succeeds → event is written, yet onDropped({reason:'timeout'}) already fired. Callbacks wired to metrics/alarms see
  phantom drops.
  2. Over-reporting — both attempts time out → one message yields three onDropped calls ([timeout, timeout, retry-failed]).

  To fix: make the timeout reject like any other failure (e.g. a typed TimeoutError) and move the reportDropped call into the post-retry loop in sendBatch, so a drop is reported
  exactly once after the retry is exhausted — classifying a terminal TimeoutError as 'timeout' and anything else as 'retry-failed'.

  Test gap: the current timeout test (sender.test.ts) asserts dropped.some(d => d.reason === 'timeout'), which passes regardless of count or duplication. Tighten it to assert the
  full dropped array (exactly one entry), and add a happy-path test — attempt 1 times out, retry succeeds → zero drops. (I confirmed locally that against the current code those
  assertions fail with 3 drops / 1 phantom drop respectively, so they pin the bug correctly.)
```

src/memory/strands/factory.ts
aidandaly24: Done in b4994b9 — `extraction: true` now passes the boolean straight through to the store's `extraction` field, so the MemoryManager resolves its own default trigger (`IntervalTrigger`) via its internal event import. We only build an `ExtractionConfig` for the explicit `{ cadence }`/`{ filter }` case, and `AgentCoreBatchTrigger` is reserved for that. Agreed the turn-based default is the right "just give me defaults" behavior since it syncs with the rest of Strands.

src/memory/strands/sender.ts
aidandaly24: Fixed in b4994b9. `addMessages` now throws (an `AggregateError`) on any `createEvent` failure, so the coordinator rolls back its high-water mark and re-fires the batch. I also removed the sender's own one-retry/`onDropped`/timeout-drop layer entirely — confirmed against `coordinator.ts` that it owns retry + backoff + repeated-failure logging, so that layer was duplicating it (and was the source of the #13 phantom-drop bug).

src/memory/strands/sender.ts
aidandaly24: Good catch, thank you for confirming the restore behavior. Fixed in b4994b9 by anchoring the `clientToken` on a run-unique id minted at sender construction (`crypto.randomUUID()` by default, overridable via `runId`) instead of `sessionId` — so a restored session that replays seq 0.. gets a different run segment and never collides. This is the interim mitigation you suggested in the next comment. Re: the dedup window — AgentCore's `clientToken` idempotency is bounded, not indefinite, which is fine here since we only need it to absorb a coordinator re-fire within a run. I'd happily switch to stable per-message UUIDs once Strands exposes them; the token derivation is a one-line change.

aidandaly24: Done exactly this in b4994b9 — a run-unique id is generated at `AgentCoreEventSender` construction and replaces `sessionId` in the token. Looking forward to swapping it for real per-message UUIDs when they land.

src/memory/strands/_strands-memory-types.ts
aidandaly24: Thanks — flipped in b4994b9. Deleted the vendored mirror and now import the interface directly from `@strands-agents/sdk` (bumped to >=1.5.0 peer / ^1.6.0 dev), and internalized the real `MessageAddedEvent` so `AgentCoreBatchTrigger` no longer needs the injected event.

src/memory/strands/store.ts
aidandaly24: Done in b4994b9 — the store-provided keys are now `_id`, `_score`, `_namespaces` (and `_createdAt`, per your other comment), underscore-prefixed so they can't collide with user-defined metadata.

src/memory/strands/batch-trigger.ts
aidandaly24: Done in b4994b9 — the batch trigger now counts only messages where `isUserOrAssistantWithText` is true (skipping tool-only/empty turns), matching the coordinator's per-turn write accounting. Added a test for it.

src/memory/strands/types.ts
aidandaly24: Done in b4994b9 — `MetadataProvider` now returns `Record<string, JSONValue>` and the sender maps each value to AgentCore's `{ stringValue }` shape internally (stringifying non-strings via `JSON.stringify`).

src/memory/strands/factory.ts
aidandaly24: You were right — the filter wasn't being threaded. Fixed in b4994b9: `AgentCoreExtractionConfig` now takes an optional `filter`, and the factory forwards it into the store's `ExtractionConfig.filter`. Added a test asserting it reaches the writable store.

src/memory/integrations/strands/store.ts
aidandaly24: Done in b4994b9 — entries now carry `_createdAt` (ISO string from `MemoryRecordSummary.createdAt`) alongside the other underscored metadata keys.

src/memory/integrations/strands/store.ts
aidandaly24: Done in b4994b9 — the store constructor now logs a warning when an extraction config is present but `writable` is false, so it isn't silently dropped. Added tests for both the warn and the no-warn (recall-only) cases.

src/memory/strands/factory.ts
aidandaly24: Done in b4994b9 — removed the factory's `ensureUniqueNames` check and deferred to the MemoryManager constructor, which already throws on duplicate store names. Dropped the now-redundant factory test.

src/memory/strands/batch-trigger.ts
aidandaly24: Fixed in b4994b9 — switched to `Buffer.byteLength(text, 'utf8')` so the `maxBytes` cap reflects the actual UTF-8 byte count AgentCore receives, not the JS string length.

src/memory/strands/sender.ts
aidandaly24: Great analysis — fixed in b4994b9, by removing the whole timeout/retry/drop layer rather than patching it. The sender now just sends and throws on failure (an `AggregateError`), and the `ExtractionCoordinator` owns retry + backoff. That eliminates both the phantom-drop and over-reporting modes you described, since there's no longer an `onDropped` callback or a timer-driven `reportDropped`. A per-request timeout, if wanted, is configured on the AWS client passed in. The old timeout test was removed with the code; the new sender tests assert the throw-on-failure / whole-batch-attempted behavior.

aidandaly24:
@jona62 Done in c9dceab — moved the module to `src/memory/integrations/strands`, matching both the Python SDK (`bedrock_agentcore.memory.integrations.strands`) and the sibling `browser`/`code-interpreter` Strands integrations in this package (both already live under `integrations/strands`). Good catch — memory was the odd one out with the flat layout. The public export subpath stays `experimental/memory/strands` to match those siblings' public paths; only the dist target moved.

aidandaly24:
**Namespace contract hardening (c9dceab)** — beyond the folder move, this commit closes the namespace silent-failure surface, which I verified against the live service:

- Records are stored under the **fully-resolved** concrete path (the service substitutes `{actorId}`/`{sessionId}`/`{memoryStrategyId}` at *extraction* time); `retrieveMemoryRecords` matches the query as a **prefix** and **rejects `{`/`}`** in a namespace.
- So an unresolved placeholder (e.g. a `/strategies/{memoryStrategyId}/...` template copied from other docs into a CLI-provisioned setup) can never match. The store now **throws at construction** with a clear message instead of failing opaquely — only `{actorId}`/`{sessionId}` are resolved client-side.
- Added validation for empty `memoryId`/`actorId`/`sessionId`/`namespace`, `minScore` ∈ [0,1], and `maxSearchResults ≥ 1` (all previously silent).
- Tests: 19 negative-space unit tests + two live integ tests — one that **asserts records land under exactly the `{actorId}`-resolved namespace the store queried** (the contract), and one proving `{sessionId}` namespaces are **per-session** (session B sees none of session A's records). Full integ suite 7/7 green, no orphaned resources.
- README + docs/MEMORY.md now document the provision-vs-query matching rule for AgentCore CLI users.

src/memory/integrations/strands/batch-trigger.ts
```diff
@@ -0,0 +1,103 @@
...
+  /** Fire after this many messages accumulate since the last fire. Default 10. */
+  messageCount?: number
+  /** Fire once accumulated message text reaches this many bytes (UTF-8). Default: unset (off). */
+  maxBytes?: number
+  /** Fire at most this long after the first un-fired message, in ms. Default 5000. `0` disables the timer. */
+  maxDelayMs?: number
+}
+
+/**
+ * Optional write-cadence trigger that fires by message count, accumulated content size, or wall-clock
```
jariy17: nit: Too many comments. Slim it down

src/memory/integrations/strands/batch-trigger.ts
```diff
@@ -0,0 +1,103 @@
...
+ * come. It is NOT needed for the common per-invocation-`flush()` pattern (flush force-drains regardless
+ * of trigger), and it is NOT the write-cost lever — payload batching in the sender (one `createEvent`
+ * per flush) reduces API calls under *any* trigger. `extraction: true` deliberately uses the Strands
+ * default; reach for this only when you need time/size-based cadence.
+ *
+ * All cadence state lives in a closure created in {@link attach}, never in instance fields, so one
+ * trigger instance attached to multiple stores keeps independent counters (mirrors the SDK's
+ * `IntervalTrigger`). The pending timer is cleared on every fire so it never strands the event loop.
+ */
+export class AgentCoreBatchTrigger extends ExtractionTrigger {
```
jariy17: I thought we are removing this Trigger and letting Strands make a Generic Batch Trigger

jariy17: Also, strand's support multiple Triggers as indicated [here](https://github.com/strands-agents/harness-sdk/blob/4173b73d0c97c16be2683a9e710de5513f6ad7aa/strands-ts/src/memory/extraction/types.ts#L129). Therefore, this compound trigger should be divided into multiple triggers (TimedTrigger, FixedSizeTrigger, MessageCountTrigger).

src/memory/integrations/strands/factory.ts
```diff
@@ -0,0 +1,291 @@
+import { BedrockAgentCoreClient } from '@aws-sdk/client-bedrock-agentcore'
```
jariy17, line 1: ---
  Spent some time on this and I think the move is to push most of the factory into the store, then keep a thin sliver of the factory for the one thing only it can do.

  What moves into the store:

  - Constructor takes a discriminated union for the read target — Subtree = { namespacePath } (required), Exact = { namespace }. Drops the readMode flag and the ReadMode type, and the store mirrors the retrieveMemoryRecords API directly instead of re-deriving the field name at call
  time.
  - Name-defaulting moves in. slugifyNamespace runs in the constructor so a store always names itself. A direct new AgentCoreMemoryStore() then works without the factory.
  - Caller passes the subtree parent explicitly instead of inferring it. That deletes commonParent.

  What gets dropped:

  - The singular createAgentCoreMemoryStore — zero added capability, callers can use the plural with [0].
  - extraction.namespace becomes obsolete. It only existed to pick which store in the list was writable, and now the caller sets writable: true on the store they want. That boolean already builds the sender, enables addMessages, and carries the extraction config.

  The catch — why I would NOT fully delete the factory:

  I checked upstream MemoryManager and there is no setting that limits writes to one store. Extraction fans out — every store with an extraction config gets scheduled independently (coordinator.schedule(store) per store). The manager is namespace-blind to AgentCore, so it wont
  dedupe the duplicate createEvents. That means the "exactly one writable/extraction-enabled store" rule can only be enforced on our side, at construction. The store cant see its siblings, and the manager wont do it.

  So the plan:

  Shrink the factory to ~15 lines whose only job is: build the shared client once, construct the stores, then assertWritableTopology(stores). The rule is at most one writable always (createEvent is namespace-free, so multiple writers send duplicate events), exactly one if you want
  extraction to run, and zero is fine for recall-only. Ill also export assertWritableTopology standalone as an escape hatch for people who build stores by hand with new.

  Net: the store mirrors the API and stands on its own, and the factory drops from ~130 lines to ~15 doing the one job nothing else can.

  ---

aidandaly24: Took this refactor — pushed across `7e1065c`, `2fe8cdb`, `b9b65b8` (plus follow-up hardening). What landed:

- **Store mirrors the API.** The read target is now a `{ namespace } | { namespacePath }` discriminated union (`types.ts` `AgentCoreReadTarget`); the `readMode` flag and `ReadMode` type are gone from the store. `search()` spreads the matched arm straight onto `RetrieveMemoryRecords`, so the field name *is* the read mode — no flag to desync.
- **Store stands alone.** `slugifyNamespace` moved into the constructor (self-naming), identity is flat (`{ memoryId, actorId, sessionId, client?, ... }` — no nested bundle), and the constructor builds its own client if none is passed. `new AgentCoreMemoryStore({ memoryId, actorId, sessionId, namespace })` now works with no factory. `writable` defaults to `false` (recall-safe).
- **`extraction.namespace` → per-namespace `writable: true`.** The writer is the namespace flagged `writable` (else the first when extraction is on). An explicit `writable: false` is honored, and `assertWritableTopology` throws on >1 writable.
- **`assertWritableTopology` added and exported** — at-most-one writable (createEvent is namespace-free, so two writers duplicate events and nothing upstream dedupes them), exactly one when extraction runs. Exported so hand-built (`new`) multi-store setups can self-guard.
- **Singular `createAgentCoreMemoryStore` dropped.** Researched the precedent and agreed it adds nothing the now-flat constructor doesn't: creating one store maps exactly to constructing one instance (so it should be the constructor, not a `plural()[0]` alias), and the framework's own `BedrockKnowledgeBaseStore` ships a constructor-only single-store path. Single store = `new AgentCoreMemoryStore({...})`; the plural factory is kept only for the multi-store case where it does real work (shared client + topology enforcement).

src/memory/integrations/strands/batch-trigger.ts
aidandaly24: Removed `AgentCoreBatchTrigger` entirely (`9048c61`) — which resolves all three notes here at once:

- **"Too many comments"** — moot, the file is gone.
- **"Letting Strands make a generic batch trigger"** — agreed; per the team discussion this is generic memory-cadence logic that belongs upstream, not in an AgentCore-specific package. It was unreleased, so removing it costs no one a public capability. The `cadence` field stays and is typed as the generic `ExtractionTrigger | ExtractionTrigger[]`, so it accepts any Strands trigger (`IntervalTrigger`/`InvocationTrigger`/an array to compose) — and a future upstream generic trigger drops straight in.
- **"Split the compound trigger into Timed/FixedSize/MessageCount"** — with the compound trigger gone there's nothing to split; composition is delegated to Strands' trigger array.

If a concrete customer need for time/size-based cadence comes up, the place to add it is upstream in the Strands repo (or back here if upstream declines), rather than re-introducing the AgentCore-specific trigger speculatively.

src/memory/integrations/strands/factory.ts
aidandaly24: Update on `commonParent`: on reflection you were right — deleted it (`5297160`). It only ran in subtree mode when `parentNamespace` was omitted, and subtree reads a single parent path anyway, so inferring it from a list of child namespaces was clunkier than just naming the parent (our own integ test already passed `parentNamespace` explicitly — nothing actually relied on the inference). Subtree mode now requires an explicit `parentNamespace` and throws a clear error if it's missing.

src/memory/integrations/strands/logger.ts
```diff
@@ -0,0 +1,37 @@
...
+ * warn/error to console), with an injection hook so callers can route logs through Pino/Winston/etc.
+ *
+ * Strands exports `configureLogging` and the `Logger` type but not its global `logger` instance, so an
+ * external package cannot import that instance. We therefore own a console-backed default here that
+ * matches the SDK's shape; if Strands later exposes its instance, this can delegate to it without
+ * touching the call sites (`logger.warn(...)`).
+ */
+
+export interface Logger {
+  debug(...args: unknown[]): void
```
jariy17, line 12: I would shift this under [_utils/](https://github.com/aws/bedrock-agentcore-sdk-typescript/blob/main/src/_utils/)

jariy17 (approved):

PR #186, opened 2026-06-09
fix: add package root entrypoint

xianzuyang9-blip:
Fixes #185.

The package root export points to `dist/src/index.js`, but the repository did not have `src/index.ts`, so `npm run build` produced subpath entrypoints like `runtime` while leaving the root export missing.

This adds a root entrypoint that exposes the existing public modules as namespaces (`identity`, `runtime`, `browser`, and `codeInterpreter`). Using namespace exports avoids collisions between shared names such as `SessionInfo`, `DEFAULT_TIMEOUT`, and `WebSocketConnection` while making the existing root export resolve to real JS and type files.

Verification:
- Before the change: `npm run build` completed, but `dist/src/index.js` was missing while `dist/src/runtime/index.js` existed.
- After the change: `npm run build` completes and emits `dist/src/index.js`.
- Packed and installed the package into a clean consumer; verified `dist/src/index.js` and `dist/src/index.d.ts` exist and `import('bedrock-agentcore')` exposes `runtime`, `identity`, `browser`, and `codeInterpreter`.
- `npm test`: 16 test files / 398 tests passed.

Note: the local pre-commit hook ran tests and lint successfully, then stopped on an existing Prettier warning in `tests_integ/otel-no-user-content.test.ts`, which this PR does not touch. I committed with `--no-verify` to keep the change scoped.

PR #190, opened 2026-06-13
Fix missing package root export

### LINKED ISSUE #185: package root export points to missing dist/src/index.js
## Summary

`bedrock-agentcore@0.2.4` publishes the package root at `dist/src/index.js`, but the npm package does not include `dist/src/index.js` or `dist/src/index.d.ts`.

Other subpath entrypoints such as `./runtime` and `./identity` are present, so this looks like a missing root build artifact or stale root export.

## Reproduction

```sh
tmp="$(mktemp -d)"
cd "$tmp"
npm init -y >/dev/null
npm install --ignore-scripts bedrock-agentcore@0.2.4

node - <<'NODE'
const fs = require('node:fs');
const path = require('node:path');
const root = path.join(process.cwd(), 'node_modules/bedrock-agentcore');
const pkg = JSON.parse(fs.readFileSync(path.join(root, 'package.json'), 'utf8'));

console.log({ main: pkg.main, rootExport: pkg.exports['.'] });
for (const file of [
  'dist/src/index.js',
  'dist/src/index.d.ts',
  'dist/src/runtime/index.js',
  'dist/src/identity/index.js',
]) {
  console.log(file, fs.existsSync(path.join(root, file)));
}
NODE

node -e "import('bedrock-agentcore')"
```

## Actual behavior

Root import fails with `ERR_MODULE_NOT_FOUND` because `dist/src/index.js` is absent.

## Expected behavior

The package root should import successfully, or the root export and `main` should point to an included file.

Environment used to reproduce: macOS, Node v24.10.0, npm 11.6.0.

alejandrorivas-pixel:
## Summary
- add the missing package root entry at `src/index.ts`
- expose the primary SDK classes and identity helpers from `bedrock-agentcore`
- keep subpath exports unchanged

Fixes #185.

## Validation
- `npm run build`
- `npm run type-check`
- `npm test` (398 tests passed)
- `npm pack --pack-destination /tmp/bedrock-agentcore-pack --json`
- installed `/tmp/bedrock-agentcore-pack/bedrock-agentcore-0.2.4.tgz` into a temporary project and verified `await import('bedrock-agentcore')` exposes `BedrockAgentCoreApp`, `RuntimeClient`, `withAccessToken`, `withApiKey`, `Browser`, and `CodeInterpreter`
- `npx prettier --check src/index.ts`

Note: the local pre-commit hook also ran `npm test` successfully, then stopped on an existing Prettier issue in `tests_integ/otel-no-user-content.test.ts`, which is unrelated to this PR.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,0 +1,25 @@
+/**
+ * Bedrock AgentCore SDK
+ */
+
+export { BedrockAgentCoreApp, RuntimeClient } from './runtime/index.js'
+export { withAccessToken, withApiKey } from './identity/index.js'
+export { Browser } from './tools/browser/index.js'
+export { CodeInterpreter } from './tools/code-interpreter/index.js'
+
+export type {
+  BedrockAgentCoreAppConfig,
+  ContentTypeParserConfig,
+  Handler,
+  RequestContext,
+  HealthStatus,
+  HealthCheckResponse,
+  AsyncTaskInfo,
+  AsyncTaskStatus,
+  RuntimeClientConfig,
+  GenerateWsConnectionParams,
+  GeneratePresignedUrlParams,
+  GenerateWsConnectionOAuthParams,
+  WebSocketConnection,
+  ParsedRuntimeArn,
+} from './runtime/index.js'


jariy17:
Thanks for the fix. Consolidating #185 into #186, which uses module namespace exports to avoid collisions and expose the existing core APIs consistently. All three implementations passed clean package-install and TypeScript consumer checks on current `main`; #186 also passed all 767 unit tests, lint, and type checks. Closing this as a duplicate so the fix has one review thread.

PR #193, opened 2026-06-15
feat(sandbox): add AgentCore CodeInterpreter Strands Sandbox implementation

gautamsirdeshmukh:
## Description

Adds an `AgentCoreSandbox` implementation of the Strands `Sandbox` interface, which serves as a thin layer over AgentCore CodeInterpreter client calls.

## Related Issues

- https://github.com/strands-agents/sdk-typescript/pull/1090/

## Documentation PR

Not yet

## Type of Change

New feature

## Testing

How have you tested the change?

- [x] I ran `npm run check`

## Checklist
- [ ] I have updated the documentation accordingly

----


agent-of-mkmeral:
## Review + E2E test of `AgentCoreSandbox`

Reviewed all 4 changed files in full and **ran the implementation end-to-end against a real AgentCore CodeInterpreter session** (`aws.codeinterpreter.v1`, us-west-2). Overall this is clean, well-documented work — the `Sandbox` contract is implemented faithfully and the native-tool mapping (vs. faking file I/O through shell) is the right call.

### ✅ E2E results — 12/12 passed against live AgentCore
Started a real session and exercised every operation through the public `Sandbox` API:

| Test | Result |
|---|---|
| `execute('echo …')` | ✅ stdout + exitCode 0 |
| `execute(env: {MYVAR})` → shell `export` prefix | ✅ `val=secret123` |
| `execute(cwd: '/tmp')` → `cd` prefix | ✅ |
| non-zero exit (`exit 3`) | ✅ exitCode 3 |
| `executeCode('print(6*7)', 'python3')` | ✅ `42` |
| `writeFile` + `readText` text roundtrip | ✅ |
| `writeFile` + `readFile` **binary roundtrip** `[0,1,2,255,254,128,7]` | ✅ byte-exact |
| `listFiles('.')` | ✅ finds written files |
| `removeFile` | ✅ |
| `readFile` on missing path | ✅ throws |
| `executeCode(…, 'ruby')` | ✅ throws "does not support language" |
| `executeStreaming` yields chunk + result | ✅ |

The headline benefit — **binary-safe file transfer** via the native `readFiles`/`writeFiles` tools — works correctly: a 7-byte buffer with high/control bytes roundtripped byte-exact. The 27 mocked unit tests, lint, and `tsc` (type-check) all pass too.

### 🔴 One thing to flag before merge: build dependency
`agentcore-sandbox.ts` imports from `@strands-agents/sdk/sandbox`, which is **not present in the currently-published `@strands-agents/sdk@0.6.0`** on npm (it has no `./sandbox` export). The base `Sandbox` interface was added in [sdk-typescript#1090](https://github.com/strands-agents/sdk-typescript/pull/1090) (merged) — but note that repo is now **archived** and the SDK moved to [`strands-agents/harness-sdk`](https://github.com/strands-agents/harness-sdk). I had to build the SDK from `harness-sdk` main and link it locally to get a clean `tsc`/test run; against published 0.6.0 the build fails with `TS2307: Cannot find module '@strands-agents/sdk/sandbox'`.

👉 Worth confirming the release path: this PR can't go green in CI / be consumed by users until an SDK version that exports `./sandbox` (and an `ExecuteOptions.env` field — see below) is published. Recommend keeping it draft and pinning the `@strands-agents/sdk` peer/dev dep to that release once it's out.

### 🟡 Minor notes (non-blocking)
1. **`env`/`cwd` on `executeCode`** — these are intentionally *not* applied to code execution (kernel has no shell), which the TSDoc explains well. Just note this diverges from the built-in `PosixShellSandbox` (Docker/SSH), where `executeCodeStreaming` pipes code through `executeStreaming` and therefore *does* honor `env`/`cwd`. The harness-sdk `ExecuteOptions.env` doc says "custom `Sandbox` implementations must handle env explicitly or it has no effect," so this is compliant — but a one-line note in the class doc that env/cwd silently no-op for `executeCode` would save a caller a debugging session.
2. **`listFiles` `isDir`** — best-effort via trailing-slash inference; when the backend omits the slash, `isDir` is reported `false` rather than `undefined`. Documented in the code. In my live run `listFiles('.')` returned plain names with `isDir:false` for real directories like `node_modules`/`run` — so directory detection effectively doesn't work against the live backend today. The `FileInfo.isDir` contract allows `undefined`; consider whether `false` ("definitely a file") is misleading here vs. `undefined` ("unknown").
3. **stderr/exitCode** come only from `structuredContent`; text content blocks only feed stdout. Correct for `executeCommand`; just flagging that `executeCode` errors surface via `isError`→exitCode=1 rather than a populated `stderr`.

### Verdict
Implementation is solid and verified working end-to-end. The only hard blocker is the unpublished/relocated SDK `./sandbox` dependency — everything else is polish. Nice work. 👍

*(Reviewed & e2e-tested by an automated Strands agent on behalf of @mkmeral.)*

gautamsirdeshmukh:
Will update NPM version before merge - seeking reviews from AgentCore SDK specialists

PR #199, opened 2026-06-25
fix(runtime/shell): make reconnect spec-conformant

padmak30:
- Drive reconnect from the socket 'close' event so connection recovery no
      longer depends on an active `for await` loop; concurrent observers share
      one in-flight attempt via _reconnectPromise.
- send()/resize() self-heal: await an in-flight reconnect and validate the
      real socket readyState, throwing a descriptive Error instead of leaking
      the raw ws "readyState 3" error.
- Keepalive treats "no Pong within ~60s" as a dead connection and
      terminates so reconnect engages, instead of sitting stale on a dead socket.
- bytesDropped is per-disconnect (assigned, not accumulated) and read from
      the single reconnection confirmation frame during the handshake.
- Wait for the authoritative 'close' before reconnecting when a socket
      'error' wakes the read loop early, so terminal codes (4000 kicked /
      1003 text-frames) are not pre-empted by a spurious reconnect; widen the
      join-guard to the 'connecting' phase to avoid double-arming.
    
    Adds unit regression coverage for the error-then-close races, keepalive
    dead-detection, and send()-after-drop self-healing.

### DIFF
--- a/.github/workflows/integration-testing.yml
+++ b/.github/workflows/integration-testing.yml
@@ -148,7 +148,10 @@ jobs:
           OAUTH2_PROVIDER_NAME: ${{ secrets.OAUTH2_PROVIDER_NAME }}
           OAUTH2_SCOPES: ${{ secrets.OAUTH2_SCOPES }}
           APIKEY_PROVIDER_NAME: ${{ secrets.APIKEY_PROVIDER_NAME }}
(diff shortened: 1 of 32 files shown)


aidandaly24 (approved):
LGTM thanks for making the change don't see any issues

Issue #66, opened 2026-02-07
BedrockAgentCoreApp throws FastifyError (invalid payload type) when invocation handler returns async generator with Accept: text/event-stream


stewartmoreland:
**Describe the bug**

When the invocation handler is an async generator (or a sync function that returns an async generator) and the client sends `Accept: text/event-stream`, the request completes with HTTP 500. The runtime calls `reply.send(payload)` with an object somewhere, and Fastify throws: "Attempted to send payload of invalid type 'object'. Expected a string or Buffer." (FST_ERR_REP_INVALID_PAYLOAD_TYPE).

**To Reproduce**

1. Install package: `yarn add bedrock-agentcore zod` (or npm equivalent).
2. Create a server using the code sample below and run it.
3. Send `POST /invocations` with header `Accept: text/event-stream` and body `{ "prompt": "hello" }`.
4. See error (500 and FastifyError in logs).

**Expected behavior**

The runtime should stream SSE events from the async generator without calling `reply.send()` with an object. The client should receive `text/event-stream` chunks from `reply.sse.send(...)`.

**Code Sample**

```typescript
import { BedrockAgentCoreApp } from 'bedrock-agentcore/runtime';
import { z } from 'zod';

const app = new BedrockAgentCoreApp({
  invocationHandler: {
    requestSchema: z.object({ prompt: z.string() }),
    process(request, ctx) {
      async function* stream() {
        yield JSON.stringify({ type: 'meta', sessionId: ctx.sessionId });
      }
      return stream();
    },
  },
});

app.run();
```

**Error Output**

```
FastifyError: Attempted to send payload of invalid type 'object'. Expected a string or Buffer.
    at onSendEnd (node_modules/fastify/lib/reply.js:652:11)
    at onSendHook (node_modules/fastify/lib/reply.js:539:5)
    at Reply.send (node_modules/fastify/lib/reply.js:210:3)
    at BedrockAgentCoreApp._handleInvocation (node_modules/bedrock-agentcore/src/runtime/app.ts:403:31)
    at Object.sseHandler (node_modules/@fastify/sse/index.js:437:37)
    ...
"msg":"Promise errored, but reply.sent = true was set"
```

**Environment (please complete the following information):**

- OS: macOS 15.7.3 (24G419)
- Node.js version: Tested with 22.15.1 and 24.13.0 — same result on both
- Package version: bedrock-agentcore@0.2.0
[...]

Issue #42, opened 2026-01-05
Module Resolution Error when Running With TSX


joshkotrous:
**Describe the bug**

**To Reproduce**

Run a typescript file using `npx tsx` that utilizes browser tools from the agent core sdk

```
node:internal/modules/esm/resolve:304
  return new ERR_PACKAGE_PATH_NOT_EXPORTED(
         ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './browser/vercel-ai' is not defined by "exports" in /Users/josh-pensar/repositories/console/nod
e_modules/bedrock-agentcore/package.json
    at exportsNotFound (node:internal/modules/esm/resolve:304:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:594:13)
    at resolveExports (node:internal/modules/cjs/loader:634:36)
    at Module._findPath (node:internal/modules/cjs/loader:724:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:1211:27)
    at nextResolveSimple (/Users/josh-pensar/repositories/console/node_modules/tsx/dist/register-D46fvsV_.cjs:4:1004)
    at /Users/josh-pensar/repositories/console/node_modules/tsx/dist/register-D46fvsV_.cjs:3:2630
    at /Users/josh-pensar/repositories/console/node_modules/tsx/dist/register-D46fvsV_.cjs:3:1542
    at resolveTsPaths (/Users/josh-pensar/repositories/console/node_modules/tsx/dist/register-D46fvsV_.cjs:4:760)
    at /Users/josh-pensar/repositories/console/node_modules/tsx/dist/register-D46fvsV_.cjs:4:1102 {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}

Node.js v20.18.1

[process exited]

```

Issue #172, opened 2026-05-15
BedrockAgentCoreApp: createRequire() for @fastify/sse breaks esbuild bundling


TheHolyWaffle:
## Problem

`BedrockAgentCoreApp` from `bedrock-agentcore/runtime` uses `createRequire()` to load `@fastify/sse` and `@fastify/websocket`:

```js
// dist/src/runtime/app.js
import { createRequire } from 'module';
const require = createRequire(import.meta.url);
const fastifySse = require('@fastify/sse');
const fastifyWebsocket = require('@fastify/websocket');
```

This pattern is opaque to esbuild's static analysis ([evanw/esbuild#1828](https://github.com/evanw/esbuild/issues/1828)). When bundling an agent into a single file with esbuild — which is the standard approach for minimal Docker images on AgentCore — `@fastify/sse` and `@fastify/websocket` are not included in the output. The container then crashes on startup:

```
Error: Cannot find module '@fastify/sse'
Require stack:
- /home/nonroot/index.js
```

## Environment

- `bedrock-agentcore`: 0.2.4
- esbuild config: `{ bundle: true, platform: 'node', format: 'esm' }`
- Docker image: distroless Node.js (only `index.js` copied, no `node_modules`)

## Suggested fix

Both `@fastify/sse` (since v4.x) and `@fastify/websocket` support ESM exports. Replace the `createRequire` pattern with direct ESM imports:

```diff
- import { createRequire } from 'module';
- const require = createRequire(import.meta.url);
- const fastifySse = require('@fastify/sse');
- const fastifyWebsocket = require('@fastify/websocket');
+ import fastifySse from '@fastify/sse';
+ import fastifyWebsocket from '@fastify/websocket';
```

This makes the dependencies visible to bundlers (esbuild, Rollup, webpack) and eliminates the runtime `createRequire` workaround entirely.

## Workarounds

Until this is fixed upstream, consumers can either:

1. Mark `@fastify/sse` and `@fastify/websocket` as `external` in esbuild and copy them into the Docker image
2. Drop the single-file bundle approach and vendor `node_modules/` instead

## Related

- #61 (Add dual module support)
- [evanw/esbuild#1828](https://github.com/evanw/esbuild/issues/1828) (`createRequire` not supported by esbuild)

PR #211, opened 2026-07-14
fix(runtime): bump @fastify/sse to 0.6.0 to stop it corrupting non-streaming error responses

### LINKED ISSUE #212: BedrockAgentCoreApp throws FST_ERR_REP_INVALID_PAYLOAD_TYPE when a non-streaming handler throws and the client sent Accept: text/event-stream
**Describe the bug**

When a client sends `Accept: text/event-stream` on a `{ sse: true }` invocation route, but the `invocationHandler.process` throws *before* ever streaming anything (i.e. it's a plain async function, not an async generator, or an async generator that throws before its first `yield`), the request fails with a confusing `FST_ERR_REP_INVALID_PAYLOAD_TYPE` instead of the expected HTTP 500 with the real error message.

This happens because `@fastify/sse@0.4.0` (the version currently pinned in `package.json`) eagerly calls `reply.raw.setHeader('Content-Type', 'text/event-stream')` as soon as the `Accept` header admits SSE — before the route handler runs at all, regardless of whether the handler ends up streaming. When `_handleInvocation`'s catch block then calls `reply.status(500).send({ error: errorMessage })`, Fastify refuses to JSON-serialize the plain object against a response whose content-type is already pinned to a non-JSON value, and throws `FST_ERR_REP_INVALID_PAYLOAD_TYPE` instead. The original error message is lost — the caller only sees the secondary serialization crash.

This is a distinct scenario from #66 / the fix in #118: that fix (correctly) guards the case where SSE streaming has already flushed headers via `writeHead(200)` (`reply.raw.headersSent === true`). Here, `headersSent` is still `false` — only the header *value* has been staged via `setHeader`, not actually sent — so that guard doesn't trigger, and the plain non-streaming `else` branch calls `reply.send()` straight into the corrupted content-type.

We hit this in production running an agent on Bedrock AgentCore Runtime: the invocation failed with a generic 500 and no indication of the real cause (in our case, a downstream database error), because the real error was swallowed by this secondary Fastify crash.

**To Reproduce**

1. Install package: `npm install bedrock-agentcore`.
2. Create a server using the code sample below and run it.
3. Send `POST /invocations` with header `Accept: text/event-stream` and a session id header, to a handler that throws synchronously (not an async generator).
4. See a 500 response, but the response body/logs show `FST_ERR_REP_INVALID_PAYLOAD_TYPE` ("Attempted to send payload of invalid type 'object'. Expected a string or Buffer.") instead of the handler's actual error message.

**Expected behavior**

The response should be a normal HTTP 500 with a JSON body `{ "error": "<the handler's actual error message>" }` — exactly as it already is when the client does *not* send `Accept: text/event-stream`.

**Code Sample**

```typescript
import { BedrockAgentCoreApp } from 'bedrock-agentcore/runtime';

const app = new BedrockAgentCoreApp({
  invocationHandler: {
    process: async (_request, _context) => {
      throw new Error('downstream dependency failed');
    },
  },
});

app.run();
```

```bash
curl -X POST http://localhost:8080/invocations \
  -H 'Content-Type: application/json' \
  -H 'Accept: text/event-stream' \
  -H 'x-amzn-bedrock-agentcore-runtime-session-id: test-session' \
  -d '{}'
```

**Error Output**

```
FastifyError: Attempted to send payload of invalid type 'object'. Expected a string or Buffer.
    at onSendEnd (node_modules/fastify/lib/reply.js:674:11)
    at onSendHook (node_modules/fastify/lib/reply.js:561:5)
    at Reply.send (node_modules/fastify/lib/reply.js:232:3)
    at BedrockAgentCoreApp._handleInvocation (node_modules/bedrock-agentcore/dist/src/runtime/app.js:371:41)
    at async Object.sseHandler (node_modules/@fastify/sse/index.js:437:15)
{"code":"FST_ERR_REP_INVALID_PAYLOAD_TYPE","statusCode":500}
```

**Environment (please complete the following information):**

- OS: macOS 14 / Amazon Linux (AWS Bedrock AgentCore Runtime container)
- Node.js version: 22.x
- Package version: 0.3.0 and 0.4.0 (confirmed via source diff — `src/runtime/app.ts` is byte-identical between the two)
- AWS SDK version: n/a (client-side bug, no AWS SDK call involved)

**Additional context**

`@fastify/sse@0.5.0` ([fastify/sse#42](https://github.com/fastify/sse/pull/42), resolving [fastify/sse#16](https://github.com/fastify/sse/issues/16)) fixes the root cause upstream: SSE response headers are now committed lazily, only when `SSEContext.sendHeaders()` is actually invoked on the first real `send()`/`stream()` call. For `{ sse: true }` ("legacy") routes, a handler that throws (or otherwise never streams) now falls through to Fastify's normal JSON serialization untouched — no changes needed to this repo's own `src/runtime/app.ts`.

**Possible Solution**

Bump the `@fastify/sse` dependency from `^0.4.0` to `^0.5.0` in `package.json`. I've verified this fixes the issue with an integration test (real Fastify + `@fastify/sse`, no mocking) that fails with the exact error above against `0.4.0` and passes against `0.5.0`. PR incoming.

hagaym1:
## Motivation

Bedrock AgentCore Runtime invocations that send `Accept: text/event-stream` on a `{ sse: true }` route — valid whenever a client doesn't know in advance whether the handler will stream — get a confusing `FST_ERR_REP_INVALID_PAYLOAD_TYPE` instead of the real error whenever the handler throws before it ever streams. `@fastify/sse@0.4.0` eagerly stages the SSE `Content-Type` header the moment `Accept` admits SSE, before the handler even runs, so a non-streaming error response gets serialized against an already-non-JSON content type and Fastify rejects it. The caller never learns why their request actually failed.

This is distinct from the fix in #118 (resolving #66): that one guards errors that occur after streaming has genuinely started and headers were flushed (`headersSent === true`). Here `headersSent` is still `false` — only the header *value* had been staged — so that guard doesn't apply, and the non-streaming branch calls `reply.send()` straight into the corrupted content type.

`@fastify/sse@0.5.0` fixed this at the source ([fastify/sse#42](https://github.com/fastify/sse/pull/42), resolving [fastify/sse#16](https://github.com/fastify/sse/issues/16)): it defers writing SSE headers until the handler actually calls `send()`. A handler that never streams now falls through to Fastify's normal JSON serialization, exactly as if `sse` had never been set. No changes to this SDK's own runtime code are needed — its control flow was already correct.

This targets `^0.6.0`, the current release, rather than `^0.5.0`: 0.6.0 keeps the deferred `sendHeaders()` behaviour and adds only an optional status-code argument to it ([fastify/sse#47](https://github.com/fastify/sse/pull/47)), which this SDK does not use.

Resolves: #212

## Verification

The regression test lives in the unit suite (`src/runtime/__tests__/app-sse-errors.test.ts`). It does not mock `@fastify/sse`, so it exercises the real plugin's `Content-Type` timing — `src/runtime/__tests__/app.test.ts` mocks the plugin and cannot cover this path. A companion test asserts SSE streaming is unaffected.

Run both ways on Node 22: the error test **fails** on `@fastify/sse@0.4.0` with a `{}` body (the masked error), and **passes** on `0.6.0`, returning HTTP 500 with `{ error: 'Sync handler blew up' }`. `npm test` (619 tests), `npm run lint`, `format:check`, `type-check` and `type-check:tests` all pass locally.

I diffed the published type surface between 0.4.0 and 0.6.0, since `SSESource` is re-exported through `RequestContext`: `SSESource` itself is unchanged. The route option widened (`sse?: boolean` → `sse?: SSERouteOptions`, still accepting `true`), and `sendHeaders()` gained an optional status-code argument. 0.6.0 also applies a stricter `Accept` gate to `sse: true` routes, which `app.ts` already handles — every use of `reply.sse` there is guarded by `if (reply.sse)`.

One item worth a maintainer's eye: `@fastify/sse` ≥ 0.5.0 declares `engines.node >= 20.20.2`, while this package declares `>= 20.0.0`. Installs on Node 20.0–20.20.1 would see an `EBADENGINE` warning (CI runs 22.x). Glad to align `engines` here if you'd prefer.

## Public API Changes

No public API changes. `{ sse: true }` routes behave identically for every case that already worked; a previously-crashing case (a non-streaming throw with `Accept: text/event-stream`) now returns the correct HTTP 500 with the real error message.

----

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "@aws-sdk/credential-providers": "^3.996.0",
         "@aws-sdk/protocol-http": "^3.370.0",
         "@aws-sdk/signature-v4": "^3.370.0",
-        "@fastify/sse": "^0.4.0",
+        "@fastify/sse": "^0.6.0",
         "@fastify/websocket": "^11.0.1",
         "@smithy/protocol-http": "^5.3.5",
         "@smithy/signature-v4": "^5.3.5",
@@ -2317,19 +2317,36 @@
       }
     },
     "node_modules/@fastify/sse": {
-      "version": "0.4.0",
-      "resolved": "https://registry.npmjs.org/@fastify/sse/-/sse-0.4.0.tgz",
-      "integrity": "sha512-bBV96iT2kHEw6h3i8IMkZGaqA7Gk81ugUzTNctXuE6N2BEC/qBnUuzlD/O17V43OkJP73h0/kf3Bp/asXlSuFA==",
+      "version": "0.6.0",
+      "resolved": "https://registry.npmjs.org/@fastify/sse/-/sse-0.6.0.tgz",
+      "integrity": "sha512-+5g8zBbSN8ioKEfll2p6TXe75peq3uJ3/z44SakK5gNr9u6YqbOAHfTSRfXfvnUFzG97+3E055l360ruzsvgcQ==",
       "license": "MIT",
       "dependencies": {
-        "fastify-plugin": "^5.0.0"
+        "@fastify/error": "^4.2.0",
+        "fastify-plugin": "^6.0.0"
       },
       "engines": {
-        "node": ">=20"
+        "node": ">=20.20.2"
       },
       "peerDependencies": {
         "fastify": "^5.x"
       }
+    },
+    "node_modules/@fastify/sse/node_modules/fastify-plugin": {
+      "version": "6.0.0",
+      "resolved": "https://registry.npmjs.org/fastify-plugin/-/fastify-plugin-6.0.0.tgz",
+      "integrity": "sha512-fZOty7z3O7vOliF6d8bHE3wiEh1KcNnKEQensSgTk9C1DvN6nRLS++XVd86v33Hw/8u9Un8A1zDrQ8ujcQDHEg==",
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/fastify"
+        },
+        {
+          "type": "opencollective",
+          "url": "https://opencollective.com/fastify"
+        }
+      ],
+      "license": "MIT"
     },
     "node_modules/@fastify/websocket": {
       "version": "11.2.0",
--- a/package.json
+++ b/package.json
@@ -148,7 +148,7 @@
     "@aws-sdk/credential-providers": "^3.996.0",
     "@aws-sdk/protocol-http": "^3.370.0",
     "@aws-sdk/signature-v4": "^3.370.0",
-    "@fastify/sse": "^0.4.0",
+    "@fastify/sse": "^0.6.0",
     "@fastify/websocket": "^11.0.1",
     "@smithy/protocol-http": "^5.3.5",
     "@smithy/signature-v4": "^5.3.5",
--- a/src/runtime/__tests__/app-sse-errors.test.ts
+++ b/src/runtime/__tests__/app-sse-errors.test.ts
@@ -1,0 +1,52 @@
+import { describe, it, expect } from 'vitest'
+import request from 'supertest'
+import type { InvocationHandler } from '../types.js'
+import { BedrockAgentCoreApp } from '../app.js'
+
+// This suite deliberately does NOT mock @fastify/sse: it exercises the real
+// plugin, because the behaviour under test is the plugin's own Content-Type
+// timing. app.test.ts mocks the plugin and therefore cannot cover it.
+const buildApp = async (handler: InvocationHandler) => {
+  const app = new BedrockAgentCoreApp({ invocationHandler: { process: handler } })
+  const fastify = (app as any)._app
+  await (app as any)._registerPlugins()
+  ;(app as any)._setupRoutes()
+  await fastify.ready()
+  return fastify
+}
+
+describe('BedrockAgentCoreApp invocation errors with the real @fastify/sse plugin', () => {
+  it('returns HTTP 500 with a JSON body when a handler throws before streaming and the client sent Accept: text/event-stream', async () => {
+    const fastify = await buildApp(async () => {
+      throw new Error('Sync handler blew up')
+    })
+
+    const res = await request(fastify.server)
+      .post('/invocations')
+      .set('x-amzn-bedrock-agentcore-runtime-session-id', 'throw-session-sse-accept')
+      .set('Accept', 'text/event-stream')
+      .send({})
+      .expect(500)
+
+    expect(res.body).toEqual({ error: 'Sync handler blew up' })
+  })
+
+  it('still streams SSE when the handler yields', async () => {
+    const fastify = await buildApp(async function* () {
+      yield 'first'
+      yield 'second'
+    })
+
+    await request(fastify.server)
+      .post('/invocations')
+      .set('x-amzn-bedrock-agentcore-runtime-session-id', 'stream-session')
+      .set('Accept', 'text/event-stream')
+      .send({})
+      .expect('Content-Type', 'text/event-stream')
+      .expect(200)
+      .expect((res) => {
+        expect(res.text).toContain('data: first')
+        expect(res.text).toContain('data: second')
+      })
+  })
+})


hagaym1:
This PR and its companion issue #212 have been open for about a week. CI is currently sitting in the first-time-contributor approval gate (the workflows show `action_required` / `waiting`), so the integration tests haven't been able to run on the change yet — a maintainer clicking "Approve and run workflows" would unblock it.

The change is small and self-contained: a single `@fastify/sse` `0.4.0` → `0.5.0` bump (fixing the eager `Content-Type` write that corrupts non-streaming error responses) plus a regression test. It's green on every check that has run, and follows the same issue-first + regression-test pattern as the previously-merged #118. Happy to rebase or adjust anything if needed.

cc @notgitika — flagging in case this fell through the cracks; no rush.

hagaym1:
Quick housekeeping update, no action needed beyond whenever it suits you:

- Rebased onto current `main` and squashed to a single commit, so this now runs under the shared reusable workflows from #221 / #227 / #231 rather than the pre-migration ones.
- Retargeted the bump to `@fastify/sse@^0.6.0` (the current release) instead of `^0.5.0`. 0.6.0 keeps the deferred `sendHeaders()` fix and adds only an optional status-code argument that this SDK doesn't use.
- Re-verified both directions on Node 22: the added test fails on `0.4.0` with the masked `FST_ERR_REP_INVALID_PAYLOAD_TYPE` (empty body) and passes on `0.6.0` with the real error message. Lint, format, type-check, build, 617 unit tests and the 58 `tests_integ/runtime.test.ts` tests all pass locally.

The `Integration Tests` run is still parked in the first-time-contributor approval gate, so it can't go green by itself — but there's genuinely no urgency here, and I'm happy for this to wait for a convenient moment.

cc @JunyangHe — only because you've been in these workflows this week, so you may already know whether the rebase is enough for the checks to pick up. Please ignore if it's not a good time.

hagaym1:
@jariy17 — apologies for the direct ping, and please point me elsewhere if this isn't your area.

Flagging you specifically because you approved #118, which fixed the sibling half of this same `FST_ERR_REP_INVALID_PAYLOAD_TYPE` bug — that one for errors raised *after* SSE headers were flushed, this one for a handler that throws *before* it ever streams (so `headersSent` is still `false` and #118's guard doesn't apply).

The change is a one-line `@fastify/sse` bump plus a regression test; no runtime code in this SDK changes. Verified both directions on Node 22: the added test fails on `0.4.0` with the masked error and an empty body, and passes on `0.6.0` returning the real message. Lint, format, type-check, build, the 617 unit tests and the 58 `tests_integ/runtime.test.ts` tests all pass locally.

It's been open since 14 July with no reviewer assigned. I'd also tried to add you as a reviewer directly, but I don't have permission to on this repo. As best I can tell a single approving review is the only real gate — the `Integration Tests` run is still parked in the first-time-contributor approval gate, so it can't report on its own either way.

No urgency at all, and glad to make any changes you'd like.

hagaym1:
@Hweinstock — you were in `src/runtime/app.ts` in #243, so this sits in your path: it's a one-line dependency bump for that file's SSE plugin.

`@fastify/sse@0.4.0` pins the SSE `Content-Type` before the route handler runs. When an `/invocations` handler throws before it streams, the SDK's own `reply.status(500).send({ error })` can't serialize against it and the real error is discarded rather than hidden. Upstream fixed it in 0.5.0. The regression test fails on 0.4.0 with a `{}` body and passes on 0.6.0 with the real message. #172 is a second, unrelated report against the same dependency.

Either of these works for me:

- `PR Validation` and `Security Scanning` are sitting at `action_required` — could you approve the runs?
- Or just land `"@fastify/sse": "^0.6.0"` yourself and close this. I don't need it to be my PR.

hagaym1:
@jariy17 — you've approved most of the outside PRs here (#118, #200, #213, #243, #245), so I think this one falls to you.

It's a one-line dependency bump: `@fastify/sse` `^0.4.0` → `^0.6.0`. On 0.4.0 the plugin pins the SSE `Content-Type` before the route handler runs, so any `/invocations` handler that throws before streaming loses its real error. The regression test fails on 0.4.0 with a `{}` body and passes on 0.6.0 with the real message. #212 has the detail; #172 is a separate report against the same dependency.

I'd rather have a decision than leave this open indefinitely — either of these works for me:

- approve the two `action_required` runs and review it, or
- land the one-line bump yourself and close this.

PR #229, opened 2026-07-28
feat(runtime): add A2A protocol support

### LINKED ISSUE #228: [FEATURE] A2A protocol support in the runtime module (parity with Python SDK's serve_a2a)
**Is your feature request related to a problem? Please describe.**

AgentCore Runtime supports hosting agents on the A2A protocol path (`serverProtocol: A2A`), and the Python SDK ships first-class support for it: `serve_a2a`, `build_a2a_app`, `build_runtime_url`, and `BedrockCallContextBuilder` in `bedrock_agentcore.runtime.a2a`. The TypeScript SDK has nothing equivalent — `BedrockAgentCoreApp` covers only the HTTP protocol path.

TypeScript teams building multi-agent systems on AgentCore today have to hand-roll the entire A2A container contract themselves: the JSON-RPC endpoint at `POST /`, the agent card at `/.well-known/agent-card.json` (including v0.3/v1.0 protocol-version compatibility), the `GET /ping` health check, extraction of the AgentCore-injected headers (session id, request id, `WorkloadAccessToken`, `OAuth2CallbackUrl`), the runtime header allowlist, and the `InvokeAgentRuntime` URL construction on the client side.

A further consequence of the gap: the identity wrappers (`withAccessToken` / `withApiKey`) rely on `getContext()` for the ambient `workloadAccessToken`, which only `BedrockAgentCoreApp` populates — so today they cannot work inside an A2A agent at all.

**Describe the solution you'd like**

A `bedrock-agentcore/runtime/a2a` subpath export that implements the AgentCore Runtime A2A container contract around the official `@a2a-js/sdk` (v1.x) server handlers, mirroring the Python SDK's surface and defaults:

- `serveA2A(options)` / `buildA2AApp(options)` — analogous to `serve_a2a` / `build_a2a_app`: JSON-RPC 2.0 at `POST /`, agent card with v0.3 legacy compat, `/ping` with an optional custom handler (`'Healthy' | 'HealthyBusy'`, falling back to `Healthy` if the handler throws), injectable `TaskStore` and `ServerCallContextBuilder`, port from `PORT` env or 9000, host auto-detection (`0.0.0.0` in containers, loopback otherwise), auto-built agent card when none is provided, and `AGENTCORE_RUNTIME_URL`-based card URL rewriting.
- **Context propagation into the existing runtime context** — the AgentCore-injected headers flow into the same `runWithContext`/`getContext` used by the HTTP path (the analog of Python's `BedrockAgentCoreContext`), so the identity wrappers work unchanged inside A2A executors, plus a `ServerCallContext.state` mirror for executors that prefer `requestContext.context.state` (the analog of `BedrockCallContextBuilder`).
- `isForwardableHeader` / header allowlist — the port of `is_forwardable_header` and `RESTRICTED_HEADERS` from the Python SDK's `runtime/models.py` (notably letting `traceparent`/`baggage` through for trace propagation).
- `buildRuntimeUrl(runtimeArn, region?)` — the analog of `build_runtime_url` for A2A clients calling deployed agents through `InvokeAgentRuntime`.
- `buildAgentCard` / `withJsonRpcUrl` — card construction helpers (Python builds its card internally; exposing these fits TS ergonomics).

`@a2a-js/sdk` and `express` would be **optional peerDependencies**, following the existing `playwright`/`@strands-agents/sdk` precedent, so the main package surface is unaffected for non-A2A users.

**Proposed API**

```typescript
import { serveA2A, buildAgentCard, getContext } from 'bedrock-agentcore/runtime/a2a'
import type { AgentExecutor } from '@a2a-js/sdk/server'

class MyExecutor implements AgentExecutor {
  async execute(requestContext, eventBus) {
    const ctx = getContext() // sessionId, requestId, workloadAccessToken, headers
    // ... run any agent framework, publish A2A task events
  }
  async cancelTask(taskId, eventBus) {}
}

await serveA2A({
  executor: new MyExecutor(),
  agentCard: buildAgentCard({
    name: 'research-agent',
    description: 'Deep research specialist',
    skills: [{ id: 'main', name: 'research', description: 'Research a topic' }],
  }),
  pingHandler: () => 'Healthy',
})
```

**Describe alternatives you've considered**

- *Fastify-native implementation instead of the `@a2a-js/sdk` Express adapters.* The HTTP path uses Fastify, so this would avoid adding `express` as an (optional) peer. `@a2a-js/sdk` exposes a transport-agnostic `JsonRpcTransportHandler`, so a Fastify mount is feasible — at the cost of hand-maintaining the card endpoint, v0.3 legacy compat, and JSON-RPC error mapping that the official Express adapters already provide. The Python SDK has the same duality precedent (its `serve_a2a` builds its own Starlette app alongside the HTTP path). We'd propose starting with the official adapters as optional peers, but we're open to either direction — flagging it here explicitly for maintainer input before the PR.
- *Publishing this as a separate community package.* Works, but the Python SDK ships A2A in the SDK itself, and context propagation into the identity wrappers genuinely belongs in the SDK (it needs the internal `runWithContext`).

**Use Case**

Multi-agent systems in TypeScript on AgentCore: a lead agent on the HTTP path delegating to specialist worker agents on the A2A path over SigV4-signed `InvokeAgentRuntime` calls. We built exactly this (with the Claude Agent SDK as the agent framework) and extracted the A2A hosting layer described here from it.

**Additional context**

We have a working implementation at feature parity with the Python SDK's `runtime/a2a.py`, developed against this repo's AGENTS.md/TESTING.md conventions (co-located tests, no `any`, TSDoc on all exports, ~96% line coverage on the new module, `npm run check` green). It has been validated end to end against a live AgentCore Runtime (eu-central-1): agent card discovery, blocking `message/send`, streaming `message/stream`, and CloudWatch-verified context propagation (session/request ids visible via `getContext()` inside the executor).

Branch: `krausexb/bedrock-agentcore-sdk-typescript@feat/a2a-runtime-support`

Deliberate deviations from Python we'd document in the PR (details there): camelCase `ServerCallContext.state` keys; `AGENTCORE_RUNTIME_URL` normalized to a trailing slash on advertised cards (Python passes it through verbatim — the slashless value breaks relative agent-card resolution for standards-compliant A2A clients); no executor introspection for the auto-card name (Python reads `executor.agent`, a Strands-specific convention).

**Would you be willing to contribute this feature?**

- [x] Yes, I would like to contribute this feature
- [ ] No, I'm just suggesting the idea

krausexb:
## Description

Adds A2A protocol support to the runtime module, at feature parity with the Python SDK's `serve_a2a` / `build_a2a_app` / `build_runtime_url` / `BedrockCallContextBuilder` (`bedrock_agentcore.runtime.a2a`).

AgentCore Runtime can host agents on the A2A protocol path, but this SDK only covers the HTTP path — TypeScript teams hand-roll the A2A container contract (JSON-RPC endpoint, agent card with v0.3/v1.0 compat, health check, AgentCore header extraction, header allowlist, runtime URL construction) and each re-discovers its non-obvious failure modes. It also means the identity wrappers can't work inside A2A agents: `withAccessToken`/`withApiKey` read the ambient `workloadAccessToken` via `getContext()`, which only `BedrockAgentCoreApp` populates today.

The new `bedrock-agentcore/runtime/a2a` subpath wraps the official `@a2a-js/sdk` (v1.x) Express server handlers:

```typescript
import { serveA2A, buildAgentCard, getContext } from 'bedrock-agentcore/runtime/a2a'

await serveA2A({
  executor: myExecutor, // any @a2a-js/sdk AgentExecutor
  agentCard: buildAgentCard({
    name: 'research-agent',
    description: 'Deep research specialist',
    skills: [{ id: 'main', name: 'research', description: 'Research a topic' }],
  }),
})
```

`serveA2A` serves JSON-RPC 2.0 at `POST /`, the agent card at `/.well-known/agent-card.json` (v1.0 + v0.3 legacy compat), and `GET /ping` with an optional `pingHandler` (`'Healthy' | 'HealthyBusy'`, degrading to `Healthy` if the handler throws). Defaults mirror Python: port from `A2A_PORT` env or 9000 (warning when it resolves to
anything else), host `0.0.0.0` inside containers (`/.dockerenv` or `DOCKER_CONTAINER`) and loopback otherwise, auto-built card when omitted, card URLs rewritten to `AGENTCORE_RUNTIME_URL` when deployed. `buildA2AApp` returns the Express app without binding for embedding/testing; `TaskStore` and `ServerCallContextBuilder` are injectable.

The key design decision: the AgentCore-injected headers (session id, request id, `WorkloadAccessToken`, `OAuth2CallbackUrl`) propagate into the **same** `runWithContext`/`getContext` request context used by the HTTP path — so the identity wrappers work unchanged inside A2A executors, with zero changes to the identity module. This is the TS analog of Python's `BedrockAgentCoreContext` ContextVars. The same fields are mirrored into `ServerCallContext.state` (the `BedrockCallContextBuilder` analog) for executors that prefer `requestContext.context.state`. Forwarded caller headers follow the runtime header allowlist ported from Python's `runtime/models.py` (`isForwardableHeader`; `traceparent`/`baggage` pass through for trace propagation).

`buildRuntimeUrl(runtimeArn, region?)` (the `build_runtime_url` analog) gives A2A clients the SigV4-signed `InvokeAgentRuntime` URL for calling deployed agents.

`@a2a-js/sdk` and `express` are **optional peerDependencies** following the existing `playwright`/`@strands-agents/sdk` precedent; the main entry points are unaffected for non-A2A users. Express (via the official `@a2a-js/sdk/server/express` adapters) rather than Fastify was chosen to reuse the upstream-maintained card endpoint, legacy compat, and JSON-RPC error mapping; a Fastify-native mount over `JsonRpcTransportHandler` is feasible if you'd rather avoid the express peer — happy to discuss (Python has the same two-stack precedent: `serve_a2a` builds its own Starlette app beside the FastAPI HTTP path).

**Deliberate deviations from the Python implementation:**

- `ServerCallContext.state` keys are camelCase (`requestId`, `workloadAccessToken`, …) per this repo's conventions, not snake_case, and the duplicated `bedrock_request_id` key is not carried over.
- `AGENTCORE_RUNTIME_URL` is normalized to a trailing slash on advertised cards (Python passes it verbatim). The platform injects the value slashless, and A2A clients resolve `/.well-known/agent-card.json` *relative* to the card URL — WHATWG resolution drops the last segment of a slashless base, so a verbatim value breaks card discovery for standards-compliant clients. Found in live validation (below).
- No executor introspection for the auto-built card's name/description — Python reads `executor.agent`, a Strands-specific convention with no generic TS equivalent. The auto-card otherwise matches Python (name `agent`, version `0.1.0`, `main` skill tagged `main`).
- `sessionId` is omitted from `ServerCallContext.state` when the header is absent (Python sets `session_id: None`).

Not in scope: Python's baggage/routing-experiment OTEL machinery (depends on Python-SDK-internal config-bundle modules with no TS counterpart) and dual a2a-sdk major-version support (the `@a2a-js/sdk` v1 legacy-compat layer covers v0.3 clients). The HTTP path's header filtering (Authorization + Custom-* only) is narrower than both Python and this A2A path; aligning it is a behavior change for existing users, left to a separate discussion.

## Related Issues

Resolves #228

## Documentation PR

No agent-docs PR yet — happy to raise one once the API is settled. In-repo docs updated: `src/runtime/README.md` (A2A section incl. an AgentCore deployment guide: `ENV DOCKER_CONTAINER=1` since the platform doesn't create `/.dockerenv`, `serverProtocol: A2A`, and the `bedrock-agentcore:GetAgentCard` IAM permission), README feature list, AGENTS.md directory structure.

## Type of Change

New feature

## Testing

- [x] I ran `npm run check`

29 unit tests in `src/runtime/a2a/__tests__/` (~96% line / 100% function coverage on the module), covering the served contract end to end over real HTTP sockets: card auto-build and URL rewrite, port/host resolution, ping semantics including the throwing-handler fallback, context propagation into both `getContext()` and `ServerCallContext.state`, custom `TaskStore`/`ServerCallContextBuilder` injection, and listen-failure rejection (`EADDRINUSE`).

Also validated end to end against a live AgentCore Runtime (eu-central-1, `serverProtocol: A2A`, image installing the `npm pack` tarball to exercise the published exports map): agent card discovery, blocking `message/send`, and streaming `message/stream` through SigV4-signed `InvokeAgentRuntime`, with CloudWatch-verified session/request ids from `getContext()` inside the executor. That validation surfaced the two trailing-slash issues fixed in this branch.

Cross-ecosystem interop with the Python SDK verified in both directions (strands-agents 1.50.2 + a2a-sdk 0.3.26, the default `bedrock-agentcore[a2a]` pin): a Python a2a-sdk 0.3 client against a `serveA2A` worker (exercises the server-side v0.3 legacy-compat layer this PR enables), and a TS v1 client against a Python `serve_a2a` + `StrandsA2AExecutor` worker.

## Checklist

----

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,10 @@
       "import": "./dist/src/runtime/index.js",
       "types": "./dist/src/runtime/index.d.ts"
     },
+    "./runtime/a2a": {
+      "import": "./dist/src/runtime/a2a/index.js",
+      "types": "./dist/src/runtime/a2a/index.d.ts"
+    },
     "./identity": {
       "import": "./dist/src/identity/index.js",
       "types": "./dist/src/identity/index.d.ts"
@@ -96,6 +100,7 @@
   "author": "Amazon Web Services",
   "license": "Apache-2.0",
   "devDependencies": {
+    "@a2a-js/sdk": "^1.0.0",
     "@ai-sdk/amazon-bedrock": "^4.0.0-beta.67",
     "@aws-sdk/client-cognito-identity-provider": "^3.957.0",
     "@eslint/js": "^10.0.1",
@@ -118,6 +123,7 @@
     "ai": "^6.0.0-beta.114",
     "eslint": "^10.0.3",
     "eslint-plugin-tsdoc": "^0.5.0",
+    "express": "^5.2.1",
     "husky": "^9.1.7",
     "jsdom": "^29.0.0",
     "playwright": "^1.56.1",
@@ -158,13 +164,21 @@
     "zod": "^4.1.13"
   },
   "peerDependencies": {
+    "@a2a-js/sdk": ">=1.0.0",
     "@strands-agents/sdk": ">=1.5.0",
     "ai": ">=6.0.0-beta",
+    "express": ">=4.21.2",
     "playwright": ">=1.56.0",
     "react": ">=18.0.0",
     "react-dom": ">=18.0.0"
   },
   "peerDependenciesMeta": {
+    "@a2a-js/sdk": {
+      "optional": true
+    },
+    "express": {
+      "optional": true
+    },
     "ai": {
       "optional": true
     },
--- a/src/runtime/a2a/__tests__/agent-card.test.ts
+++ b/src/runtime/a2a/__tests__/agent-card.test.ts
@@ -1,0 +1,74 @@
+import { describe, it, expect, afterEach, vi } from 'vitest'
+import { buildAgentCard, withJsonRpcUrl } from '../agent-card.js'
+
+describe('buildAgentCard', () => {
+  afterEach(() => {
+    vi.unstubAllEnvs()
+  })
+
+  it('builds a card with v1.0 and legacy v0.3 JSONRPC interfaces', () => {
+    const card = buildAgentCard({ name: 'my-agent', description: 'demo', url: 'http://localhost:9000/' })
+
+    expect(card.name).toBe('my-agent')
+    expect(card.description).toBe('demo')
+    expect(card.capabilities?.streaming).toBe(true)
+    const bindings = card.supportedInterfaces.map((i) => ({ url: i.url, protocolVersion: i.protocolVersion }))
+    expect(bindings).toEqual([
+      { url: 'http://localhost:9000/', protocolVersion: '1.0' },
+      { url: 'http://localhost:9000/', protocolVersion: '0.3' },
+    ])
+  })
+
+  it('resolves the URL from AGENTCORE_RUNTIME_URL when no url is given', () => {
+    vi.stubEnv('AGENTCORE_RUNTIME_URL', 'https://runtime.example.com/invocations')
+    const card = buildAgentCard({ name: 'deployed', description: 'x' })
+    expect(card.supportedInterfaces[0]!.url).toBe('https://runtime.example.com/invocations')
+  })
+
+  it('falls back to localhost with the given port', () => {
+    const card = buildAgentCard({ name: 'local', description: 'x', port: 3005 })
+    expect(card.supportedInterfaces[0]!.url).toBe('http://localhost:3005/')
+  })
+
+  it('normalizes partial skills to complete AgentSkill objects', () => {
+    const card = buildAgentCard({
+      name: 'skilled',
+      description: 'x',
+      skills: [{ id: 'main', name: 'main', description: 'primary skill' }],
+    })
+    expect(card.skills).toEqual([
+      {
+        id: 'main',
+        name: 'main',
+        description: 'primary skill',
+        tags: [],
+        examples: [],
+        inputModes: [],
+        outputModes: [],
+        securityRequirements: [],
+      },
+    ])
+  })
+})
+
+describe('withJsonRpcUrl', () => {
+  it('rewrites every JSONRPC interface URL', () => {
+    const card = buildAgentCard({ name: 'stale', description: 'x', url: 'http://stale:9000/' })
+    const updated = withJsonRpcUrl(card, 'https://fresh.example.com/')
+
+    expect(updated.supportedInterfaces.map((i) => i.url)).toEqual([
+      'https://fresh.example.com/',
+      'https://fresh.example.com/',
+    ])
+    // Original card is not mutated
+    expect(card.supportedInterfaces[0]!.url).toBe('http://stale:9000/')
+  })
+
+  it('appends a JSONRPC interface when the card has none', () => {
+    const card = { ...buildAgentCard({ name: 'no-rpc', description: 'x' }), supportedInterfaces: [] }
+    const updated = withJsonRpcUrl(card, 'https://fresh.example.com/')
+    expect(updated.supportedInterfaces).toEqual([
+      { url: 'https://fresh.example.com/', protocolBinding: 'JSONRPC', tenant: '', protocolVersion: '1.0' },
+    ])
+  })
+})
--- a/src/runtime/a2a/__tests__/agent-card.test.ts
+++ b/src/runtime/a2a/__tests__/agent-card.test.ts
@@ -20,9 +20,18 @@ describe('buildAgentCard', () => {
   })
 
   it('resolves the URL from AGENTCORE_RUNTIME_URL when no url is given', () => {
+    vi.stubEnv('AGENTCORE_RUNTIME_URL', 'https://runtime.example.com/invocations/')
+    const card = buildAgentCard({ name: 'deployed', description: 'x' })
+    expect(card.supportedInterfaces[0]!.url).toBe('https://runtime.example.com/invocations/')
+  })
+
+  it('appends a trailing slash to a slashless AGENTCORE_RUNTIME_URL', () => {
+    // The platform injects the env value without a trailing slash; clients
+    // resolving the well-known card path relative to a slashless URL lose
+    // the final path segment.
     vi.stubEnv('AGENTCORE_RUNTIME_URL', 'https://runtime.example.com/invocations')
     const card = buildAgentCard({ name: 'deployed', description: 'x' })
-    expect(card.supportedInterfaces[0]!.url).toBe('https://runtime.example.com/invocations')
+    expect(card.supportedInterfaces[0]!.url).toBe('https://runtime.example.com/invocations/')
   })
 
   it('falls back to localhost with the given port', () => {
--- a/src/runtime/a2a/agent-card.ts
+++ b/src/runtime/a2a/agent-card.ts
@@ -1,0 +1,115 @@
+/**
+ * A2A agent card construction with AgentCore Runtime URL resolution.
+ */
+
+import { duplicateInterfacesForLegacy } from '@a2a-js/sdk/compat/v0_3'
+import type { AgentCard, AgentSkill } from '@a2a-js/sdk'
+
+const AGENTCORE_RUNTIME_URL_ENV = 'AGENTCORE_RUNTIME_URL'
+
+/**
+ * Parameters for building an A2A agent card.
+ */
+export interface AgentCardParams {
+  /**
+   * Human-readable agent name.
+   */
+  name: string
+
+  /**
+   * Short description of what the agent does.
+   */
+  description: string
+
+  /**
+   * Card version string, defaults to '1.0.0'.
+   */
+  version?: string
+
+  /**
+   * Skills advertised on the card; id, name, and description are required, remaining fields default to empty.
+   */
+  skills?: Array<Pick<AgentSkill, 'id' | 'name' | 'description'> & Partial<AgentSkill>>
+
+  /**
+   * Base URL where the agent is reachable; defaults to the AGENTCORE_RUNTIME_URL env var, then localhost.
+   */
+  url?: string
+
+  /**
+   * Port used for the localhost fallback URL, defaults to 9000.
+   */
+  port?: number
+}
+
+/**
+ * Builds an A2A v1.0 AgentCard with a legacy v0.3 interface mirror.
+ *
+ * The service URL resolution order is: explicit `url` param, the
+ * `AGENTCORE_RUNTIME_URL` environment variable (set when deployed on
+ * AgentCore Runtime), then `http://localhost:{port}/`. Both a v1.0 and a
+ * v0.3 JSONRPC interface are declared because AgentCore's documented A2A
+ * shape and the Python A2A ecosystem still speak v0.3, while `@a2a-js/sdk`
+ * v1 clients use the v1.0 methods — the server's legacy compat layer
+ * routes both.
+ *
+ * @param params - Name, description, and optional skills/url/port
+ * @returns A complete AgentCard
+ */
+export function buildAgentCard(params: AgentCardParams): AgentCard {
+  const url = params.url ?? process.env[AGENTCORE_RUNTIME_URL_ENV] ?? `http://localhost:${params.port ?? 9000}/`
+
+  return {
+    name: params.name,
+    description: params.description,
+    version: params.version ?? '1.0.0',
+    supportedInterfaces: duplicateInterfacesForLegacy(
+      [{ url, protocolBinding: 'JSONRPC', tenant: '', protocolVersion: '1.0' }],
+      ['JSONRPC']
+    ),
+    provider: undefined,
+    capabilities: {
+      streaming: true,
+      pushNotifications: false,
+      extensions: [],
+      extendedAgentCard: false,
+    },
+    securitySchemes: {},
+    securityRequirements: [],
+    defaultInputModes: ['text/plain'],
+    defaultOutputModes: ['text/plain'],
+    skills: (params.skills ?? []).map((skill) => ({
+      id: skill.id,
+      name: skill.name,
+      description: skill.description,
+      tags: skill.tags ?? [],
+      examples: skill.examples ?? [],
+      inputModes: skill.inputModes ?? [],
+      outputModes: skill.outputModes ?? [],
+      securityRequirements: skill.securityRequirements ?? [],
+    })),
+    signatures: [],
+    iconUrl: undefined,
+  }
+}
+
+/**
+ * Returns a copy of `card` with every JSONRPC interface pointing at `url`,
+ * appending one if the card has none.
+ *
+ * Used by {@link serveA2A} so a caller-provided card never advertises a
+ * stale URL when `AGENTCORE_RUNTIME_URL` is set. The runtime invocation URL
+ * serves both protocol versions, so all JSONRPC interfaces (v1.0 and the
+ * v0.3 legacy mirror) get the same value.
+ *
+ * @param card - The agent card to rewrite
+ * @param url - The URL to set on all JSONRPC interfaces
+ * @returns A new AgentCard with updated interfaces
+ */
+export function withJsonRpcUrl(card: AgentCard, url: string): AgentCard {
+  const hasJsonRpc = card.supportedInterfaces.some((i) => i.protocolBinding === 'JSONRPC')
+  const supportedInterfaces = hasJsonRpc
+    ? card.supportedInterfaces.map((i) => (i.protocolBinding === 'JSONRPC' ? { ...i, url } : i))
+    : [...card.supportedInterfaces, { url, protocolBinding: 'JSONRPC', tenant: '', protocolVersion: '1.0' }]
+  return { ...card, supportedInterfaces }
+}
--- a/src/runtime/a2a/agent-card.ts
+++ b/src/runtime/a2a/agent-card.ts
@@ -8,6 +8,24 @@ import type { AgentCard, AgentSkill } from '@a2a-js/sdk'
 const AGENTCORE_RUNTIME_URL_ENV = 'AGENTCORE_RUNTIME_URL'
 
 /**
+ * Returns the AGENTCORE_RUNTIME_URL env value with a trailing slash, or
+ * undefined when the variable is unset.
+ *
+ * The platform injects the value without a trailing slash, but A2A clients
+ * resolve the well-known agent-card path relative to the advertised URL —
+ * and WHATWG URL resolution drops the final path segment of a slashless
+ * base, breaking card discovery. Values passed explicitly by callers are
+ * not normalized.
+ */
+export function agentCoreRuntimeUrl(): string | undefined {
+  const url = process.env[AGENTCORE_RUNTIME_URL_ENV]
+  if (!url) {
+    return undefined
+  }
+  return url.endsWith('/') ? url : `${url}/`
+}
+
+/**
  * Parameters for building an A2A agent card.
  */
 export interface AgentCardParams {
@@ -57,7 +75,7 @@ export interface AgentCardParams {
  * @returns A complete AgentCard
  */
 export function buildAgentCard(params: AgentCardParams): AgentCard {
-  const url = params.url ?? process.env[AGENTCORE_RUNTIME_URL_ENV] ?? `http://localhost:${params.port ?? 9000}/`
+  const url = params.url ?? agentCoreRuntimeUrl() ?? `http://localhost:${params.port ?? 9000}/`
 
   return {
     name: params.name,
--- a/src/runtime/a2a/app.ts
+++ b/src/runtime/a2a/app.ts
@@ -26,12 +26,10 @@ import type { AgentCard } from '@a2a-js/sdk'
 import { agentCardHandler, jsonRpcHandler, UserBuilder } from '@a2a-js/sdk/server/express'
 import type { FastifyBaseLogger } from 'fastify'
 
-import { buildAgentCard, withJsonRpcUrl } from './agent-card.js'
+import { agentCoreRuntimeUrl, buildAgentCard, withJsonRpcUrl } from './agent-card.js'
 import { extractA2AContext } from './headers.js'
 import { getContext, runWithContext } from '../context.js'
 import type { HealthStatus, RequestContext } from '../types.js'
-
-const AGENTCORE_RUNTIME_URL_ENV = 'AGENTCORE_RUNTIME_URL'
 
 /**
  * Options for {@link serveA2A} and {@link buildA2AApp}.
@@ -233,7 +231,7 @@ export const bedrockCallContextBuilder: ServerCallContextBuilder = (options) => 
 }
 
 function resolveAgentCard(provided: AgentCard | undefined, port: number): AgentCard {
-  const runtimeUrl = process.env[AGENTCORE_RUNTIME_URL_ENV]
+  const runtimeUrl = agentCoreRuntimeUrl()
   if (!provided) {
     // The generic fallback card advertises the actual listen port;
     // AGENTCORE_RUNTIME_URL takes precedence inside buildAgentCard. Name,
--- a/src/runtime/a2a/app.ts
+++ b/src/runtime/a2a/app.ts
@@ -1,0 +1,270 @@
+/**
+ * A2A protocol server for AWS Bedrock AgentCore Runtime.
+ *
+ * Implements the AgentCore Runtime A2A container contract around the
+ * `@a2a-js/sdk` Express handlers:
+ *
+ * - JSON-RPC 2.0 endpoint at `POST /`
+ * - Agent card at `GET /.well-known/agent-card.json`
+ * - Health check at `GET /ping`
+ *
+ * AgentCore Runtime's A2A path is a transparent proxy: `InvokeAgentRuntime`
+ * payloads pass through to `POST /` unmodified, so there is no envelope to
+ * unwrap. The AgentCore-injected headers (session id, request id, workload
+ * access token, OAuth2 callback URL) are propagated into the same request
+ * context used by the HTTP protocol path, so identity wrappers such as
+ * `withApiKey` work unchanged inside A2A executors.
+ */
+
+import { existsSync } from 'fs'
+import type { Server } from 'http'
+import express from 'express'
+import type { Express, NextFunction, Request, Response } from 'express'
+import { DefaultRequestHandler, InMemoryTaskStore, ServerCallContext, STATE_HEADERS_KEY } from '@a2a-js/sdk/server'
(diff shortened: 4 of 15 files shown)


src/runtime/a2a/agent-card.ts
```diff
@@ -0,0 +1,133 @@
...
+ * Returns the AGENTCORE_RUNTIME_URL env value with a trailing slash, or
+ * undefined when the variable is unset.
+ *
+ * The platform injects the value without a trailing slash, but A2A clients
+ * resolve the well-known agent-card path relative to the advertised URL —
+ * and WHATWG URL resolution drops the final path segment of a slashless
+ * base, breaking card discovery. Values passed explicitly by callers are
+ * not normalized.
+ */
+export function agentCoreRuntimeUrl(): string | undefined {
```
aidandaly24, line 20: I wonder if this function should be moved somewhere that is more of a central util file? @jariy17 what do you think?

src/runtime/a2a/app.ts
```diff
@@ -0,0 +1,276 @@
...
+    return buildAgentCard({
+      name: 'agent',
+      description: 'A Bedrock AgentCore agent',
+      version: '0.1.0',
+      skills: [{ id: 'main', name: 'agent', description: 'A Bedrock AgentCore agent', tags: ['main'] }],
+      port,
+    })
+  }
+  return runtimeUrl ? withJsonRpcUrl(provided, runtimeUrl) : provided
+}
```
aidandaly24, line 248: Shouldn't we try to get the card from Runtime?
https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/bedrock-agentcore/command/GetAgentCardCommand/

src/runtime/a2a/headers.ts
```diff
@@ -0,0 +1,251 @@
...
+const OAUTH2_CALLBACK_HEADER = 'oauth2callbackurl'
+const AUTHORIZATION_HEADER = 'authorization'
+const CUSTOM_HEADER_PREFIX = 'x-amzn-bedrock-agentcore-runtime-custom-'
+
+/**
+ * Headers that must not be forwarded to agent code, from the AgentCore
+ * runtime header allowlist docs — the same list the Python SDK ships in
+ * `runtime/models.py`. Grouped as in the docs.
+ */
+const RESTRICTED_HEADERS: ReadonlySet<string> = new Set(
```
aidandaly24, line 59: I wonder if these should be moved out of the A2A folder as well this looks useful.

src/runtime/a2a/agent-card.ts
```diff
@@ -0,0 +1,133 @@
...
+      tags: skill.tags ?? [],
+      examples: skill.examples ?? [],
+      inputModes: skill.inputModes ?? [],
+      outputModes: skill.outputModes ?? [],
+      securityRequirements: skill.securityRequirements ?? [],
+    })),
+    signatures: [],
+    iconUrl: undefined,
+  }
+}
```
aidandaly24, line 112: Will this format ever need to be changed?

src/runtime/a2a/runtime-url.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @returns The full invocation URL with the ARN percent-encoded
+ * @throws Error when no well-formed AWS region can be determined
+ *
+ * @example
+ * ```typescript
+ * const url = buildRuntimeUrl('arn:aws:bedrock-agentcore:us-east-1:123456789012:runtime/my-agent-abc123')
+ * // 'https://bedrock-agentcore.us-east-1.amazonaws.com/runtimes/arn%3Aaws%3A.../invocations'
+ * ```
+ */
+export function buildRuntimeUrl(runtimeArn: string, region?: string): string {
```
aidandaly24, line 24: Same thing here, this seems to be more of a general runtime thing rather than A2A specific.

package.json
```diff
@@ -96,6 +100,7 @@
   "author": "Amazon Web Services",
   "license": "Apache-2.0",
   "devDependencies": {
+    "@a2a-js/sdk": "^1.0.0",
```
aidandaly24, line 103: Should we have a cap pin on this so it doesn't bump major version automatically when a new one comes out.

aidandaly24 (commented):
A few questions/suggestions below. Thanks!

src/runtime/a2a/agent-card.ts
krausexb: Happy to move it to a central place. Currently the A2A module is the only consumer, so either placement works; let me know what you and @jariy17 prefer.

One thing to keep in mind. The function is normalizing the env value to a trailing slash, which is specifically needed for A2A agent-card resolution (clients resolve /.well-known/agent-card.json relative to the URL, and WHATWG resolution drops the last segment of a slashless base). If we make it a general util in src/_utils/, I'd propose splitting it: a raw agentCoreRuntimeUrl() returning the env value verbatim, with the slash normalization staying at the card layer, ensuring no future non-A2A consumer get surprised.

Let me know your preference so that I can update the implementation.

src/runtime/a2a/app.ts
krausexb: resolveAgentCard and GetAgentCardCommand sit on opposite ends. resolveAgentCard sits on the server side and builds the card that the agent container itself serves at /.well-known/agent-card.json. Runtime's GetAgentCard API is implemented the other way around: AgentCore proxies it to this very endpoint in the container. The container is the source of truth for the card, and it learns its own public URL from the AGENTCORE_RUNTIME_URL env var the platform injects, no API call needed. 

On the client side you're right that GetAgentCardCommand is the AWS-SDK way to fetch a deployed agent's card.

src/runtime/a2a/headers.ts
krausexb: Agreed, the Python SDK has the same placement. RESTRICTED_HEADERS/is_forwardable_header live in the shared runtime/models.py, not the a2a module. I'd propose to move them to src/runtime/headers.ts and keep the re-export from the runtime/a2a subpath so the published API is unchanged. 

The re-export keeps the a2a subpath self-contained (same pattern as getContext, which also lives in shared runtime code), but if you'd rather expose header utilities only from bedrock-agentcore/runtime, or keep isForwardableHeader internal entirely and only export extractA2AContext, both are one-line changes now and much harder after release.

What do you think?

src/runtime/a2a/agent-card.ts
krausexb: The shape is the AgentCard type from @a2a-js/sdk, so it tracks the A2A spec version of the dependency. If a future SDK changes the card shape, this breaks loudly at compile time rather than drifting. Protocol-version evolution is already handled. The card advertises both a v1.0 interface and a v0.3 legacy mirror via the SDK's compat helper. The hard-coded values (streaming on, push notifications off, text/plain modes) mirror the Python SDK's card defaults. For anything beyond the exposed params, the escape hatch is passing a fully custom AgentCard to serveA2A, buildAgentCard is just a convenience. If you'd like, I can widen AgentCardParams with optional overrides (capabilities, defaultInputModes/OutputModes, securitySchemes) so little tweaks don't require hand-building the whole card.

src/runtime/a2a/runtime-url.ts
krausexb: Good catch, there's an even more concrete win here: it should reuse getDataPlaneEndpoint from src/_utils/endpoints.ts instead of hard-coding the host, which also picks up the BEDROCK_AGENTCORE_DATA_PLANE_ENDPOINT override that runtime/client.ts already honors. I'll make that change.

package.json
krausexb: The devDependency here is already major-safe, caret ranges don't cross majors, so ^1.0.0 resolves to >=1.0.0 <2.0.0 and a v2 release won't be installed automatically. The place a v2 could slip through is the peerDependency, which is currently ">=1.0.0" with no upper bound. I followed the repo's existing peer style (playwright >=1.56.0, react >=18.0.0, etc., all uncapped). Since this module builds directly on @a2a-js/sdk's Express adapters and v0.3 compat layer, I think capping that one to ">=1.0.0 <2.0.0" is justified even though it deviates from the other peers. Happy to make that change.

src/runtime/a2a/agent-card.ts
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * Returns a copy of `card` with every JSONRPC interface pointing at `url`,
+ * appending one if the card has none.
+ *
+ * Used by {@link serveA2A} so a caller-provided card never advertises a
```
Hweinstock: is there a risk this comment becomes stale? Having a function describe its consumers means we need to update this comment each time we add/change a consumer.

src/runtime/a2a/agent-card.ts
```diff
@@ -0,0 +1,133 @@
...
+  /**
+   * Port used for the localhost fallback URL, defaults to 9000.
+   */
+  port?: number
+}
+
+/**
+ * Builds an A2A v1.0 AgentCard with a legacy v0.3 interface mirror.
+ *
+ * The service URL resolution order is: explicit `url` param, the
```
Hweinstock: It feels like this comment describes the implementation exactly, meaning if we change the implementation we also must update the comment, which feels like unnecessary churn. 

I think we should reserve comments for cases where the code is unable to express something, instead of reiterating what the code is expressing. 

In this example, the fallback order is precisely described by the `??` order below, so I'm not sure a comment is necessary.

src/runtime/a2a/agent-card.ts
```diff
@@ -0,0 +1,133 @@
+/**
```
Hweinstock: I don't think comments describing a files purpose are necessary, unless the code in that file doesn't make its purpose obvious. 

In this case, the filename is `agent-card` so I don't think its necessary. I think the same idea applies to a few of these comments.

src/runtime/a2a/__tests__/agent-card.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    expect(card.capabilities?.streaming).toBe(true)
+    const bindings = card.supportedInterfaces.map((i) => ({ url: i.url, protocolVersion: i.protocolVersion }))
+    expect(bindings).toEqual([
+      { url: 'http://localhost:9000/', protocolVersion: '1.0' },
+      { url: 'http://localhost:9000/', protocolVersion: '0.3' },
+    ])
+  })
+
+  it('resolves the URL from AGENTCORE_RUNTIME_URL when no url is given', () => {
+    vi.stubEnv('AGENTCORE_RUNTIME_URL', 'https://runtime.example.com/invocations/')
```
Hweinstock, line 23: is there a reason we need to mock this rather than modifying the actual environment? If there's an opportunity to test real behavior without mocks, I think that provides stronger guarantees.

src/runtime/a2a/app.ts
```diff
@@ -0,0 +1,276 @@
...
+
+  /**
+   * Agent card to serve; auto-built when omitted, URL-rewritten to AGENTCORE_RUNTIME_URL when that env var is set.
+   */
+  agentCard?: AgentCard
+
+  /**
+   * Port to serve on; defaults to the PORT env var, or 9000 (the AgentCore A2A protocol port) when unset.
+   */
+  port?: number
```
Hweinstock, line 71: are we sure this should be configurable? I thought 9000 was part of the protocol spec: https://docs.aws.amazon.com/en_en/bedrock-agentcore/latest/devguide/runtime-a2a-protocol-contract.html#container-requirements.

src/runtime/a2a/runtime-url.ts
Hweinstock: ^ good catch, looks like https://github.com/aws/bedrock-agentcore-sdk-typescript/blob/89f6b057336b752b3ff6fe125299592e766d023f/src/_utils/endpoints.ts#L90 should be partially re-usable.

src/runtime/a2a/headers.ts
```diff
@@ -0,0 +1,251 @@
...
+}
+
+/**
+ * Raw incoming headers as provided by Node.js HTTP servers.
+ */
+export type IncomingHeaders = Record<string, string | string[] | undefined>
+
+const SESSION_HEADER = 'x-amzn-bedrock-agentcore-runtime-session-id'
+const REQUEST_ID_HEADER = 'x-amzn-bedrock-agentcore-runtime-request-id'
+const WORKLOAD_TOKEN_HEADER = 'workloadaccesstoken'
```
Hweinstock, line 44: I noticed the http agent has a fallback mechanism for this header. Is this same needed here? 

source: https://github.com/aws/bedrock-agentcore-sdk-typescript/blob/89f6b057336b752b3ff6fe125299592e766d023f/src/runtime/app.ts#L597

src/runtime/a2a/app.ts
```diff
@@ -0,0 +1,276 @@
...
+  // Bind all interfaces only where the container contract needs it; on a
+  // developer machine an A2A agent has no business listening externally.
+  const inContainer = existsSync('/.dockerenv') || Boolean(process.env.DOCKER_CONTAINER)
+  const host = options.host ?? (inContainer ? '0.0.0.0' : '127.0.0.1')
+  const agentCard = resolveAgentCard(options.agentCard, port)
+
+  const app = buildA2AApp({ ...options, agentCard, port })
+
+  return new Promise((resolve, reject) => {
+    // Express 5 invokes the listen callback with the error on bind failure
```
Hweinstock: does this also work with Express 4? I noticed the peer dependency allows this here https://github.com/krausexb/bedrock-agentcore-sdk-typescript/blob/7a3e1de0678dd5e9c4231cce56f7ba89ffd12e67/package.json#L170, but the behavior differs across express versions based on https://expressjs.com/en/guide/migrating-5/#app.listen.

src/runtime/a2a/app.ts
```diff
@@ -0,0 +1,276 @@
...
+    // (e.g. EADDRINUSE) instead of emitting an unhandled 'error' event.
+    const server = app.listen(port, host, (error?: Error) => {
+      if (error) {
+        reject(error)
+        return
+      }
+      console.log(`agent=<${agentCard.name}>, host=<${host}>, port=<${port}> | a2a server listening`)
```
Hweinstock: is it worth exposing some logging configuration similar to how the http template does this? https://github.com/aws/bedrock-agentcore-sdk-typescript/blob/89f6b057336b752b3ff6fe125299592e766d023f/src/runtime/app.ts#L290

I see a start for some of this here https://github.com/aws/bedrock-agentcore-sdk-typescript/blob/7a3e1de0678dd5e9c4231cce56f7ba89ffd12e67/src/runtime/a2a/app.ts#L256-L275

Hweinstock (commented):
Cool feature, and thanks for making the change! I'm not too familiar with the original app implementation or a2a, but I had a few questions/comments. 

I also think some more e2e-style testing could be valuable. It looks like most of this could be tested by spinning up a local server and hitting it with some test requests.

src/runtime/a2a/headers.ts
krausexb: Good catch, we indeed should read IDENTITY_WAT_HEADER first in extractA2AContext and use workloadaccesstoken as a fallback. Additionally, we should add an explicit allow for IDENTITY_WAT_HEADER in the forwarded-header path, matching app.ts:588-592.

krausexb: Fixed in https://github.com/krausexb/bedrock-agentcore-sdk-typescript/commit/a1bfb91ce2540d62ee8ae9d56e2d099039a0169b

src/runtime/a2a/app.ts
krausexb: On Express 4 the promise never settles and the 'error' event goes unhandled, so a bind failure crashes the process rather than rejecting. The callback-error signature we relied on is Express-5-only.

Fixed by settling on the server's own events, which are the one shape both majors agree on:

```
const server = app.listen(port, host)
server.once('error', reject)
server.once('listening', () => {
  server.removeListener('error', reject)
  resolve(server)
})
```

src/runtime/a2a/app.ts
krausexb: Wanted to keep the port flexible for testing, but reading PORT is dangerous since an image might have set it to a different protocol's port, which would break the contract.

Changed it so that A2A_PORT is read by default and if no value is specified default to 9000.

Also introduced warn logs if started with a port other than 9000.

src/runtime/a2a/runtime-url.ts
krausexb: Done! buildRuntimeUrl now goes through getDataPlaneEndpoint, so it also picks up the BEDROCK_AGENTCORE_DATA_PLANE_ENDPOINT override that runtime/client.ts already honours. Previously it pasted the host together itself, so setting that variable redirected the rest of the SDK but not this function.

"Partially" was the right word. Kept the region-format check, since the helper only rejects empty regions and this function derives the region from the ARN. Also had to trim a trailing slash off the override before joining, as it's returned verbatim and may or may not have one.

On @aidandaly24's original point about placement, I checked where Python puts each of these, and it splits in two:

- build_runtime_url lives in Python's runtime/a2a.py, so I've left runtime-url.ts here — moving it would actually diverge from the Python layout.
- is_forwardable_header / RESTRICTED_HEADERS do live in the shared runtime/models.py, so those moved to src/runtime/headers.ts, re-exported from the A2A subpath. Public exports are unchanged.

Let me know if this placement works for you.

src/runtime/a2a/app.ts
krausexb: Added a logger option. Startup, the off-contract port warning, the ping-handler failure and context.log inside executors all route through it, defaulting to a console-backed logger when omitted:

await serveA2A({ executor, logger: pino({ level: 'debug' }) })

I went with an injected logger rather than mirroring HTTP's logging: { enabled, options }. An options object is a recipe someone has to build a logger from. Fastify does that for the HTTP path, but there's no framework in the middle here, so we'd have to build it ourselves. Taking a finished instance avoids that and is strictly more capable, since you can hand us a pino configured however you like, or a logger you already have.

src/runtime/a2a/agent-card.ts
krausexb: Agreed, no defence for that one — and the `{@link}` made it look load-bearing on top. Rewritten to state why the rewrite exists rather than who calls it:

> A card built for local use otherwise keeps advertising a localhost URL once deployed, which breaks client discovery.

I swept the rest of the module for the same pattern. The remaining `{@link}`s describe what a type is *for* ("Options for `serveA2A`") or point at the alternative ("use `serveA2A` to also listen) — those do not go stale when a caller is added, so I left them. Shout if you would rather they went too.

src/runtime/a2a/agent-card.ts
krausexb: Right on the fallback order — dropped that paragraph, the `??` chain says it.

I kept the block itself and one paragraph. `eslint-plugin-tsdoc` and AGENTS.md require TSDoc with `@param`/`@returns` on exports, so it cannot go entirely. And the surviving paragraph explains why the card declares *both* a v1.0 and a v0.3 JSONRPC interface, which is not inferable from `duplicateInterfacesForLegacy(...)` — it comes from AgentCore's documented card shape plus the state of the Python A2A ecosystem. Without it that call looks redundant and someone deletes it.

src/runtime/a2a/agent-card.ts
krausexb: Removed from `agent-card.ts` and `runtime-url.ts` — both just restated the filename.

I read "a few of these" as *some* rather than all, so I trimmed the other two instead of deleting them. `app.ts` keeps the note that AgentCore's A2A path is a transparent proxy with no envelope to unwrap, and that the injected headers land in the same request context as the HTTP path; `headers.ts` keeps the fact that the platform injects those headers per request. Neither is discoverable from the code, and `runtime/shell/protocol.ts` sets the precedent in this repo for that kind of header. `index.ts` also keeps its header since it documents the optional peer deps. Happy to drop those two as well if you would rather have it uniform.

Your point also prompted a sweep of every comment in the module against the code it annotates, which turned up a few more of my own:

- The `A2ALogger` level methods had TSDoc that only restated their names ("Logs an error." above `error()`) — removed; `child()` keeps its doc since the bindings and return value are not obvious.
- The WHATWG trailing-slash rationale was copy-pasted in four places — now stated once on `agentCoreRuntimeUrl` with short pointers elsewhere.
- A couple of comments and one test comment that restated their own next line.

Net about 70 lines of comments gone across the module, with the *why* comments kept — Express 4/5 listen behaviour, auth termination at the proxy, ping degradation, context propagation.

krausexb:
On the e2e testing point — that is the approach the suite already takes, so worth pointing you at it: `src/runtime/a2a/__tests__/app.test.ts` binds real ephemeral sockets and drives them with real `fetch`. Nothing is mocked at the HTTP layer, and there is no mocking of the AWS SDK or of `@a2a-js/sdk` either. Card discovery, `/ping`, JSON-RPC `message/send`, `EADDRINUSE`, injected `TaskStore`/`contextBuilder` and the context propagation are all asserted against a running server.

That said, taking the comment seriously turned up six real gaps, and all six are now covered:

- workload access token arriving as `x-amz-bedrock-agentcore-identity-wat` (the agent-to-agent case)
- `A2A_PORT` honoured, generic `PORT` ignored, and the off-contract-port warning
- `BEDROCK_AGENTCORE_DATA_PLANE_ENDPOINT` override in `buildRuntimeUrl`
- `message/stream` as SSE, which every served card advertises via `capabilities.streaming`
- both protocol versions on the request path
- restricted caller headers withheld from executors, end to end

The protocol-version one was the interesting find. `message/send` and `message/stream` are **v0.3** method names that only resolve through the legacy compat layer — flipping `legacyCompat` off makes them fail with `-32601 Invalid method` and breaks 10 existing tests. So the suite was already exercising the v0.3 path throughout, and the genuinely untested path was v1.0 native (`SendMessage`). Both wire shapes are now asserted explicitly, which also documents the difference: v0.3 returns `kind: "task"` with a lowercase state, v1.0 wraps the task and reports `TASK_STATE_COMPLETED`.

I mutation-tested the new tests rather than trusting that they pass: disabling `legacyCompat` fails the v0.3 and streaming tests while the v1.0 test keeps passing, and forwarding every header fails only the header test.

Hweinstock:
Oh I see the tests now, must have missed this on the first pass.

Hweinstock (approved):
lgtm, thanks for addressing earlier comments.

PR #243, opened 2026-08-05
feat(runtime): provide tools for consuming and propagating WAT

### LINKED ISSUE #235: Port Identity WAT header propagation from Python SDK (bedrock-agentcore-sdk-python#607)
## Summary

The Python SDK is adding automatic propagation of the workload identity chain via a new
`X-Amz-Bedrock-AgentCore-Identity-WAT` header
([aws/bedrock-agentcore-sdk-python#607](https://github.com/aws/bedrock-agentcore-sdk-python/pull/607)).
This issue tracks the TypeScript counterpart, and records where a 1:1 port does **not** apply
because the two SDKs are shaped differently.

## What the Python PR does

1. Adds an `IDENTITY_WAT_HEADER = "X-Amz-Bedrock-AgentCore-Identity-WAT"` constant.
2. Allowlists it in `is_forwardable_header` as an explicit exception to the `x-amz-*` block.
3. Inbound: `_build_request_context` reads the WAT header, falling back to the legacy
   `WorkloadAccessToken` header.
4. Outbound: a botocore `before-sign` event handler injects the header on `Invoke*` operations,
   registered on the Runtime data plane client. Identity is deliberately excluded (it is the WAT
   issuer, not a consumer).
5. Adds a `@requires_wat()` decorator that reads the WAT from context and injects it into the
   decorated function.

### 1. Constant + inbound extraction — `src/runtime/app.ts` (`_extractContext`)

TypeScript has no equivalent of `models.py`'s header-constant module and no `is_forwardable_header`
helper. `_extractContext` filters inbound headers inline against a hardcoded two-case allowlist
(`authorization`, `x-amzn-bedrock-agentcore-runtime-custom-`). Python items 1, 2 and 3 therefore
collapse into a single edit here:

```ts
// third case in the header filter loop
'x-amz-bedrock-agentcore-identity-wat'

const workloadAccessToken =
  (request.headers['x-amz-bedrock-agentcore-identity-wat'] as string) ||
  (request.headers['workloadaccesstoken'] as string) ||
  undefined
```

Fastify lowercases inbound header keys, so no case-insensitivity handling is needed — the Python
casing tests have no counterpart here.

Open decision: add a separate `identityWat` field to `RequestContext` (`src/runtime/types.ts`), or
reuse the existing `workloadAccessToken` slot. Python reuses the same slot, and reuse means
`withAccessToken` / `withApiKey` (`src/identity/wrappers.ts`) keep working with no change.
Recommend reuse.

### 2. Outbound propagation

There is no botocore event system. The equivalent is a Smithy middleware:

```ts
client.middlewareStack.add(
  next => async args => {
    const wat = getContext()?.workloadAccessToken
    if (wat && HttpRequest.isInstance(args.request)) {
      args.request.headers[IDENTITY_WAT_HEADER] = wat
    }
    return next(args)
  },
  { step: 'finalizeRequest', priority: 'high', name: 'identityWatPropagation' }
)
```

Two constraints worth settling before this is written:

**There is currently no internal client to register it on.** Python hooks
`AgentCoreRuntimeClient.dp_client`, which calls `invoke_agent_runtime`. The TypeScript
`RuntimeClient` (`src/runtime/client.ts`) is WebSocket-only — SigV4 signing, presigned URLs and
shell sessions; it has no invoke path. Every `BedrockAgentCoreClient` that does get constructed in
this repo (`src/identity/client.ts`, the Strands memory factory and store, the browser and
code-interpreter clients) is either Identity — explicitly excluded — or a non-`Invoke*` operation.

So the proposal is to export `registerIdentityWatPropagation(client)` from
`src/_utils/identity-propagation.ts` for consumers to apply to their own client, and document that
it has no internal caller yet. Note that Smithy middleware is registered per-client, not per-
operation, so operation gating has to happen inside the handler rather than at registration time.

**`InvokeGateway` does not exist.** `@aws-sdk/client-bedrock-agentcore` ships
`InvokeAgentRuntimeCommand`, `InvokeAgentRuntimeCommandCommand`, `InvokeHarnessCommand`,
`InvokeBrowserCommand` and `InvokeCodeInterpreterCommand` — there is no gateway invoke API. See the
cross-SDK note at the bottom; this entry should not be ported.

The outbound hook that is genuinely useful on the TypeScript side, and has no Python analogue:
`RuntimeClient.generateWsConnection` is TypeScript's real outbound runtime invoke path, so
attaching the WAT to those headers is where this feature earns its keep.

### 3. `withWat()` HOF — `src/identity/wrappers.ts`, exported from `src/identity/index.ts`

Python's kwarg-injecting decorator becomes last-positional-arg injection, matching the two existing
wrappers:

```ts
export function withWat() {
  return <TParams extends [...unknown[], string], TReturn>(fn: (...a: TParams) => Promise<TReturn>) =>
    async (...args: InitParams<TParams>): Promise<TReturn> => {
      const wat = getContext()?.workloadAccessToken
      if (!wat) {
        throw new Error(
          'No workload access token in context. Ensure the agent is running on AgentCore runtime.'
        )
      }
      return fn(...([...args, wat] as any))
    }
}
```

No sync/async split — the existing TypeScript wrappers are async-only. The `into:` parameter is
dropped; positional injection makes it meaningless.

### 4. Tests

- `src/runtime/__tests__/app.test.ts` — header extraction and fallback precedence
- `src/identity/__tests__/wrappers.test.ts` — `withWat` injects, and throws with no WAT in context
- new `src/_utils/__tests__/identity-propagation.test.ts` — middleware injects when a WAT is
  present, no-ops when absent

Vitest, using `runWithContext` for setup. No `contextvars.Context()` equivalent is needed — the
`AsyncLocalStorage` scoping already provides isolation.

## Estimate

Roughly 4 files, ~120 lines. Two Python changes have no TypeScript counterpart: the
`is_forwardable_header` allowlist exception, and the header-casing tests. The boto3 event-handler
registration ports to Smithy middleware but has no internal caller, so it ships as an exported
helper only.

## Cross-SDK note for the Python PR

Two things worth resolving on
[#607](https://github.com/aws/bedrock-agentcore-sdk-python/pull/607) before this is ported:

- `"InvokeGateway"` in `_WAT_PROPAGATION_OPERATIONS` appears to be dead — there is no such boto3
  operation, so that handler registration never fires.
- The module docstring tells users to attach the header manually for Gateway MCP calls, since those
  go over raw HTTP with no boto3 API. That manual path is the one that actually needs the WAT, which
  makes the dead allowlist entry misleading about what is covered automatically.

Hweinstock:
### Problem 
Closes: https://github.com/aws/bedrock-agentcore-sdk-typescript/issues/235

Summary: The python sdk received some changes to make it easier to consume and propagate the WAT, we want the same here. 

### Solution
- Add `withWat` wrapper that injects WAT from context into function as last paramter (following existing pattern). 
- Add `registerIdentityWatPropagation` middleware that propagates the token on all outbound requests for a given AWS SDK client. 
- add logic to accept incoming WAT. 

### Testing/Verification 
- added unit tests that verify the flow up to the network call. 
- added integ tests the verify the whole flow.

### DIFF
--- a/src/identity/middleware.ts
+++ b/src/identity/middleware.ts
@@ -3,6 +3,7 @@ import type {
   FinalizeHandler,
   FinalizeHandlerArguments,
   FinalizeHandlerOutput,
+  HandlerExecutionContext,
   MetadataBearer,
   MiddlewareStack,
 } from '@smithy/types'
@@ -18,7 +19,7 @@ interface AwsClient {
 
 /**
  * Registers middleware that attaches the WAT header to outbound AWS SDK requests.
- * No-ops when no WAT is present in context.
+ * Only applies to Invoke* operations. No-ops when no WAT is present in context.
  *
  * @param client - Any AWS SDK v3 client instance
  *
@@ -34,9 +35,12 @@ interface AwsClient {
 export function withWatPropagation(client: AwsClient): void {
   client.middlewareStack.add(
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    (next: FinalizeHandler<any, any>): FinalizeHandler<any, any> =>
+    (next: FinalizeHandler<any, any>, context: HandlerExecutionContext): FinalizeHandler<any, any> =>
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       async (args: FinalizeHandlerArguments<any>): Promise<FinalizeHandlerOutput<any>> => {
+        if (!context.commandName?.startsWith('Invoke')) {
+          return next(args)
+        }
         const wat = getContext()?.workloadAccessToken
         if (wat && HttpRequest.isInstance(args.request)) {
           args.request.headers[IDENTITY_WAT_HEADER] = wat
--- a/src/identity/middleware.ts
+++ b/src/identity/middleware.ts
@@ -1,0 +1,48 @@
+import { HttpRequest } from '@aws-sdk/protocol-http'
+import type {
+  FinalizeHandler,
+  FinalizeHandlerArguments,
+  FinalizeHandlerOutput,
+  MetadataBearer,
+  MiddlewareStack,
+} from '@smithy/types'
+import { getContext } from '../runtime/context.js'
+import { IDENTITY_WAT_HEADER } from '../runtime/constants.js'
+
+// AWS SDK v3 does not export a generic client interface.
+// https://github.com/aws/aws-sdk-js-v3/issues/5856#issuecomment-2096950979
+interface AwsClient {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  middlewareStack: MiddlewareStack<any, MetadataBearer>
+}
+
+/**
+ * Registers middleware that attaches the WAT header to outbound AWS SDK requests.
+ * No-ops when no WAT is present in context.
+ *
+ * @param client - Any AWS SDK v3 client instance
+ *
+ * @example
+ * ```typescript
+ * import { BedrockAgentCoreClient } from '@aws-sdk/client-bedrock-agentcore'
+ * import { withWatPropagation } from 'bedrock-agentcore/identity'
+ *
+ * const client = new BedrockAgentCoreClient({ region: 'us-west-2' })
+ * withWatPropagation(client)
+ * ```
+ */
+export function withWatPropagation(client: AwsClient): void {
+  client.middlewareStack.add(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (next: FinalizeHandler<any, any>): FinalizeHandler<any, any> =>
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      async (args: FinalizeHandlerArguments<any>): Promise<FinalizeHandlerOutput<any>> => {
+        const wat = getContext()?.workloadAccessToken
+        if (wat && HttpRequest.isInstance(args.request)) {
+          args.request.headers[IDENTITY_WAT_HEADER] = wat
+        }
+        return next(args)
+      },
+    { step: 'finalizeRequest', priority: 'high', name: 'identityWatPropagation' }
+  )
+}
--- a/src/runtime/__tests__/identity-wat.test.ts
+++ b/src/runtime/__tests__/identity-wat.test.ts
@@ -1,0 +1,93 @@
+import { describe, expect, it } from 'vitest'
+import { BedrockAgentCoreApp } from '../app.js'
+import { withWAT } from '../../identity/wrappers.js'
+import type { RequestContext } from '../types.js'
+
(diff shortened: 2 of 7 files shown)


src/runtime/__tests__/identity-wat.test.ts
```diff
@@ -0,0 +1,93 @@
+import { describe, expect, it } from 'vitest'
```
Hweinstock, line 1: note: i made a separate file for this since the existing one mocks all of fastify. This file instead injects requests into fastify allowing us to get real routing, parsing, etc.

src/runtime/app.ts
```diff
@@ -575,23 +576,27 @@ export class BedrockAgentCoreApp<TSchema extends z.ZodSchema = z.ZodSchema<unkno
     // Filter headers to include only Authorization and Custom-* headers
     const filteredHeaders: Record<string, string> = {}
     for (const [key, value] of Object.entries(request.headers)) {
-      const lowerKey = key.toLowerCase()
-      const stringValue = typeof value === 'string' ? value : Array.isArray(value) ? value.join(', ') : undefined
+      const lowerCaseHeaderName = key.toLowerCase()
```
Hweinstock, line 579: renamed this because I thought they were vague.

Hweinstock:
```
Run actions/checkout@v6
Error: Refusing to check out fork pull request code from a 'pull_request_target' workflow. This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access. Fetching and executing a fork's code in that trusted context commonly leads to "pwn request" vulnerabilities. To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.
```

appears to be an access issue, but was able to verify the newly added integ tests pass locally.

src/identity/middleware.ts
```diff
@@ -0,0 +1,48 @@
...
+ * @example
+ * ```typescript
+ * import { BedrockAgentCoreClient } from '@aws-sdk/client-bedrock-agentcore'
+ * import { withWatPropagation } from 'bedrock-agentcore/identity'
+ *
+ * const client = new BedrockAgentCoreClient({ region: 'us-west-2' })
+ * withWatPropagation(client)
+ * ```
+ */
+export function withWatPropagation(client: AwsClient): void {
```
jariy17, line 35: withWatPropagation should only propagate the WAT if it's part of an allowlist: 
```
  - InvokeAgentRuntime
  - InvokeAgentRuntimeCommand
  - InvokeHarness
  - InvokeGateway
```

jariy17 (approved):
LGTM just a follow up pr on this.

Hweinstock:
discussed in person, merging with new commit then going to to change to rule to require review on latest commit.

PR #249, opened 2026-08-11
fix(runtime): update shell session wire protocol

priyasingh2111:
## Description

Simplify shell session connection and lifecycle to improve performance and align with updated server behavior.

**Changes:**
- **Faster connection** — shell is ready immediately after WebSocket opens. Removed unnecessary handshake wait that added up to 10s latency.
- **`close()` now detaches instead of terminating** — closing the session disconnects the WebSocket but leaves the shell alive on the server, enabling reconnection within the reconnect window.
- **Removed `shell.reconnected` and `shell.bytesDropped` properties** — these are no longer available. TypeScript will report a compile error if accessed.
- **Updated `onReconnect` callback signature** — now `() => void | Promise<void>` (no arguments).

Exit codes, auto-reconnect, and all other shell functionality continue to work as before.

## Type of Change

Breaking change

## Breaking Changes

- `shell.reconnected` — removed (was `boolean`)
- `shell.bytesDropped` — removed (was `number`)
- `onReconnect` callback signature: `(reconnected: boolean) => void` → `() => void | Promise<void>`
- `close()` no longer terminates the shell — it detaches, allowing reconnection

## Testing

- [x] I ran `npm run check`
- [x] `npm test` — 616 tests pass (24 files), 1.13s
- [x] `npm run lint` — clean
- [x] `npm run format:check` — clean
- [x] `npm run type-check` — clean
- [x] Integration tested against live runtime (8/8 scenarios pass)

## Checklist

---

### DIFF
--- a/src/runtime/shell/config.ts
+++ b/src/runtime/shell/config.ts
@@ -32,7 +32,6 @@ export const noopLogger: Logger = {
 export const DEFAULT_MAX_RETRIES = 5
 export const DEFAULT_BASE_DELAY = 1000 // ms
 export const DEFAULT_MAX_DELAY = 15000 // ms
-export const DEFAULT_METADATA_TIMEOUT = 10_000 // ms
 export const DEFAULT_RECONNECT_WINDOW = 900_000 // ms — ~15 min, matches server-side KARP idle timeout
 export const DEFAULT_OUTER_LOOP_DELAY = 30_000 // ms
 export const DEFAULT_KEEPALIVE_INTERVAL = 30_000 // ms — KARP idle timeout is ~60s; ping every 30s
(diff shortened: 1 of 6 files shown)


src/runtime/shell/session.ts
```diff
@@ -514,25 +475,10 @@ export class ShellSession implements AsyncIterable<ShellFrame> {
...
-      ws.terminate()
-      return
-    }
-
     const keepaliveTimer = this._startKeepalive(ws, controller.signal)
-    // Atomic promotion: all connection objects become available together.
-    this._state = { status: 'open', ws, messageIterator, keepaliveTimer, pendingFrames }
+    // Connection is ready immediately after WebSocket opens — shellId comes from 101 header.
+    // No longer blocking on the 0x03 confirmation frame.
+    this._state = { status: 'open', ws, messageIterator, keepaliveTimer }
```
aidandaly24, line 503: Could we check that the socket is still open before setting the session state here? If `close` arrives immediately after `open`, the close handler still sees `connecting` and skips reconnect, then this line records the closed socket as `open`. This regresses the close-event-driven, iterator-independent reconnect behavior added in #199. A guard on `ws.readyState` or `_closeError` here could fail the initial connection or let the existing reconnect loop retry.

src/runtime/shell/config.ts
```diff
@@ -86,8 +84,6 @@ export interface ReconnectConfig {
 
   /**
    * Optional callback invoked after each successful reconnect.
-   * Receives `reconnected: true` when the existing PTY was reattached;
-   * `false` when a fresh shell was started.
    */
-  onReconnect?: (reconnected: boolean) => void | Promise<void>
+  onReconnect?: () => void | Promise<void>
```
aidandaly24, line 89: Could we also update the `openShell()` example in `src/runtime/client.ts` with this signature change? It still uses `onReconnect: (r) => ...`, but the callback now accepts no arguments, so copied code no longer matches the public type.

Issue #182, opened 2026-06-03
Parity with Python lib


hamatoyogi:
**Is your feature request related to a problem? Please describe.**
It's pretty frustrating that the TS sdk is missing features from the Python SDK.
At the end of the day, I think most developers building AI systems and platforms benefit a lot by using Typescript.
There's already a few issues that are complaining about specific features missing.

**Describe the solution you'd like**
It would be beneficial to share the roadmap or have a few open issues for each feature so the community can maybe contribute towards this advancement. 

**Describe alternatives you've considered**
Using agentcore just via Python - for me personally, less ideal.

**Use Case**
Building robust type safe systems that can easily integrate in full stack ts projects.

**Would you be willing to contribute this feature?**

- [x] Yes, I would like to contribute this feature
- [ ] No, I'm just suggesting the idea

PR #257, opened 2026-09-04
feat(tools): add WebSearchClient for AgentCore Web Search

sundargthb:
## Description

Web search is reachable today as an AgentCore Gateway connector target that an agent calls as an MCP tool. This SDK had no way to reach it, so a TypeScript caller had to hand-roll SigV4 signing and a JSON-RPC MCP conversation before getting a single result. Python is getting the same client in aws/bedrock-agentcore-sdk-python#658, which is still open, so this surface may want to follow that review.

The scope here is deliberately narrow: a web search client plus a gateway MCP endpoint helper. It does not attempt the Gateway module the README lists as coming soon.

### Public API changes

New subpath export, `bedrock-agentcore/web-search`:

```typescript
import { WebSearchClient } from 'bedrock-agentcore/web-search'

const client = new WebSearchClient({
  region: 'us-east-1',
  gatewayId: 'my-gateway-abc123',
  targetName: 'amazon-web-search', // optional; skips a tools/list round trip
})

const response = await client.search('what shipped in node 24', {
  maxResults: 5,
  includeDomains: ['aws.amazon.com'],
})

for (const result of response.results) {
  console.log(result.title, result.url)
}

client.close()
```

`getGatewayMcpEndpoint(gatewayId, region)` is added to `src/_utils/endpoints.ts`, mirroring `get_gateway_mcp_endpoint` in the Python SDK. Gateway identifiers become the first DNS label of the host, so it rejects an ARN or a URL rather than building a bad hostname.

The transport sits behind a `WebSearchBackend` interface (`search`, `close`), and `GatewayMcpBackend` is one implementation of it. A different access path can then be added as an addition rather than a break for callers, and passing `backend` to the client is also what lets the unit tests run with neither a network nor credentials.

### Use cases

Any TypeScript agent that needs a search tool: the client returns a normalized `{ text, url, title, publishedDate }` rather than MCP content blocks, and it keeps `url` and `title` on the type because AgentCore's acceptable use terms require citations to be displayed to end users.

Pass `targetName` when you know it. Gateway prefixes every tool with the name of the target it came from (`amazon-web-search___WebSearch`), so supplying the target name derives that directly instead of spending a `tools/list` call on the first search.

Two decisions reviewers may want to make differently:

- whether this belongs at `src/tools/web-search/` or under a future gateway module.
- `eslint.config.js` needed `Response` and `RequestInit` added to the `src` globals allowlist. `no-undef` is active there with an explicit list, and `AbortSignal` was already on it.

### Breaking Changes

None. Everything is additive.

## Related Issues

Contributes to #182.

## Documentation PR

None needed. The docs are in this PR: `docs/WEB_SEARCH.md`, plus a README feature entry.

## Type of Change

New feature

## Testing

92 unit tests across the two touched areas, covering argument shaping and limits, response normalization, the SigV4 headers (including that `content-length` is not signed, since fetch sets it itself and signing it invalidates the signature), MCP session reuse under concurrency, tool discovery and its failure modes, both the `application/json` and `text/event-stream` framings, and timeouts.

`npm run lint`, `npm run format:check`, `npm run type-check:all`, `npm run build` and `npm test` (698 tests) all pass locally.

- [ ] I ran `npm run check`

Not ticked, and the reason is not this branch: `check` includes `test:coverage`, which reports 25 per-file threshold failures on an untouched `main`. This branch reports the same 25, all in files it does not touch. `tools/web-search` itself is at 98% lines and 92% branches.

No integration test is included, because `tests_integ` would need a gateway with a web search connector target in the integ account.

## Checklist

The last one is unticked because the Python counterpart is still open. Nothing in this PR depends on it, but the two surfaces should match.

---

### DIFF
--- a/src/_utils/endpoints.ts
+++ b/src/_utils/endpoints.ts
@@ -26,6 +26,38 @@ const GATEWAY_ID_PATTERN = /^[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?$/
 const REGION_PATTERN = /^[a-z0-9-]+$/
 
 /**
+ * Host suffixes an AWS endpoint can have.
+ */
+const AWS_DOMAINS = ['.amazonaws.com', '.amazonaws.com.cn', '.api.aws']
+
+/**
+ * Checks that an endpoint URL resolves to an AWS host.
+ *
+ * Requests to these endpoints are signed with the caller's credentials, including a
+ * session token, so an endpoint that came from configuration or user input gets
+ * checked before anything is signed for it. This also catches URL manipulation that
+ * slips past the gateway identifier and region patterns.
+ *
+ * @param url - The endpoint URL to check
+ * @returns The URL, unchanged
+ *
+ * @throws Error if the URL cannot be parsed or its host is not an AWS host.
+ */
+export function validateEndpointUrl(url: string): string {
+  let hostname: string
+  try {
+    hostname = new URL(url).hostname
+  } catch {
+    throw new Error(`Not a valid endpoint URL: '${url}'`)
+  }
+
+  if (!AWS_DOMAINS.some((domain) => hostname.endsWith(domain))) {
+    throw new Error(`Endpoint resolves to a non-AWS host: '${hostname}'`)
+  }
+  return url
+}
+
+/**
  * Gets the data plane endpoint for the Bedrock AgentCore service.
  *
  * The endpoint can be overridden using the BEDROCK_AGENTCORE_DATA_PLANE_ENDPOINT
@@ -71,7 +103,8 @@ export function getDataPlaneEndpoint(region: string): string {
  * gateway call is computed over.
  *
  * The endpoint can be overridden using the BEDROCK_AGENTCORE_GATEWAY_ENDPOINT
- * environment variable.
+ * environment variable. An override is returned as given, so a local MCP mock can be
+ * pointed at, while a URL built from arguments is checked to resolve to an AWS host.
  *
  * @param gatewayId - Gateway identifier, not an ARN (e.g. 'my-gateway-abc123')
  * @param region - AWS region the gateway lives in (e.g. 'us-east-1')
@@ -107,5 +140,5 @@ export function getGatewayMcpEndpoint(gatewayId: string, region: string): string
     return override
   }
 
-  return `https://${gatewayId}.gateway.bedrock-agentcore.${region}.amazonaws.com/mcp`
+  return validateEndpointUrl(`https://${gatewayId}.gateway.bedrock-agentcore.${region}.amazonaws.com/mcp`)
 }
--- a/src/_utils/endpoints.ts
+++ b/src/_utils/endpoints.ts
@@ -8,6 +8,24 @@
 const ENDPOINT_OVERRIDE_ENV = 'BEDROCK_AGENTCORE_DATA_PLANE_ENDPOINT'
 
 /**
+ * Environment variable for overriding the gateway MCP endpoint.
+ */
+const GATEWAY_ENDPOINT_OVERRIDE_ENV = 'BEDROCK_AGENTCORE_GATEWAY_ENDPOINT'
+
+/**
+ * A gateway identifier becomes the first DNS label of the endpoint host, so it is
+ * restricted to what a label allows. This rejects an ARN, which is the mistake
+ * this validation exists to catch.
+ */
+const GATEWAY_ID_PATTERN = /^[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?$/
+
+/**
+ * Region names are lowercase alphanumerics and hyphens, such as `us-east-1`.
+ * Validated because it is interpolated into a hostname.
+ */
+const REGION_PATTERN = /^[a-z0-9-]+$/
+
+/**
  * Gets the data plane endpoint for the Bedrock AgentCore service.
  *
  * The endpoint can be overridden using the BEDROCK_AGENTCORE_DATA_PLANE_ENDPOINT
@@ -44,4 +62,50 @@ export function getDataPlaneEndpoint(region: string): string {
 
   // Return standard AWS endpoint pattern
   return `https://bedrock-agentcore.${region}.amazonaws.com`
+}
+
+/**
+ * Builds the streamable HTTP MCP endpoint for an AgentCore Gateway.
+ *
+ * This is the URL an MCP client posts to, and the URL a SigV4 signature for a
+ * gateway call is computed over.
+ *
+ * The endpoint can be overridden using the BEDROCK_AGENTCORE_GATEWAY_ENDPOINT
+ * environment variable.
+ *
+ * @param gatewayId - Gateway identifier, not an ARN (e.g. 'my-gateway-abc123')
+ * @param region - AWS region the gateway lives in (e.g. 'us-east-1')
+ * @returns Full HTTPS endpoint URL, including the /mcp path
+ *
+ * @throws Error if the gateway identifier is not a valid DNS label, or if the
+ * region is empty or malformed.
+ *
+ * @example
+ * ```typescript
+ * const endpoint = getGatewayMcpEndpoint('my-gateway-abc123', 'us-east-1')
+ * // Returns: 'https://my-gateway-abc123.gateway.bedrock-agentcore.us-east-1.amazonaws.com/mcp'
+ * ```
+ */
+export function getGatewayMcpEndpoint(gatewayId: string, region: string): string {
+  if (!gatewayId || !GATEWAY_ID_PATTERN.test(gatewayId)) {
+    throw new Error(
+      `Invalid gateway identifier: '${gatewayId}'. Expected a gateway ID such as 'my-gateway-abc123', not an ARN or URL.`
+    )
+  }
+
+  if (!region || region.trim() === '') {
+    throw new Error('Region cannot be empty')
+  }
+
+  if (!REGION_PATTERN.test(region)) {
+    throw new Error(`Invalid region: '${region}'. Expected a region such as 'us-east-1'.`)
+  }
+
+  // Check for environment variable override
+  const override = process.env[GATEWAY_ENDPOINT_OVERRIDE_ENV]
+  if (override) {
+    return override
+  }
+
+  return `https://${gatewayId}.gateway.bedrock-agentcore.${region}.amazonaws.com/mcp`
 }
--- a/src/tools/web-search/client.ts
+++ b/src/tools/web-search/client.ts
@@ -12,7 +12,7 @@ import { HttpRequest } from '@smithy/protocol-http'
 import { SignatureV4 } from '@smithy/signature-v4'
 import { Sha256 } from '@aws-crypto/sha256-js'
 
-import { getGatewayMcpEndpoint } from '../../_utils/endpoints.js'
+import { getGatewayMcpEndpoint, validateEndpointUrl } from '../../_utils/endpoints.js'
 import type {
   SearchOptions,
   WebSearchBackend,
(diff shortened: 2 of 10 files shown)


src/tools/web-search/client.ts
```diff
@@ -0,0 +1,776 @@
...
+    const negotiated = reply.result?.protocolVersion
+    if (typeof negotiated === 'string' && negotiated) {
+      this.protocolVersion = negotiated
+    }
+
+    // Set before the notification is sent, because from here on every request carries
+    // mcp-protocol-version. If the notification fails the handshake did not complete,
+    // so this is rolled back rather than left claiming a session the gateway never
+    // acknowledged.
+    this.initialized = true
```
tejaskash: There is a small race here. `initialized` flips to true before the `notifications/initialized` POST on line 304 has finished, and `ensureInitialized` (line 266) checks `initialized` before it checks `initializing`. So if a second `search()` lands after the initialize reply comes back but before the notification has gone out, it skips the wait entirely and sends `tools/call` ahead of the notification. If the notification then fails, `resetSession()` runs while that second call is already in flight with the old session id.

The concurrency test does not catch this because `Promise.all` fires both calls before `initialized` flips. In an agent loop issuing staggered tool calls the window is narrow but real.

Two ways out: have `ensureInitialized` check `this.initializing` first, or keep `initialized` false until the notification returns and give `post` an explicit flag for whether to send `mcp-protocol-version`.

src/tools/web-search/client.ts
```diff
@@ -0,0 +1,776 @@
...
+
+  private async initialize(): Promise<void> {
+    const reply = await this.post({
+      jsonrpc: '2.0',
+      id: this.nextId(),
+      method: 'initialize',
+      params: {
+        protocolVersion: MCP_PROTOCOL_VERSION,
+        capabilities: {},
+        clientInfo: { name: 'bedrock-agentcore-typescript', version: MCP_PROTOCOL_VERSION },
```
tejaskash: `clientInfo.version` should be the SDK's version, not the MCP protocol version. As written, gateway logs will show this client as version `2025-06-18`, which will confuse anyone reading them. The package version is the right value here, or just drop the field.

src/tools/web-search/client.ts
```diff
@@ -0,0 +1,776 @@
...
+   */
+  private async signedHeaders(body: string, extra: Record<string, string>): Promise<Record<string, string>> {
+    const url = new URL(this.endpoint)
+    const credentials = await this.credentialsProvider()
+
+    const request = new HttpRequest({
+      protocol: url.protocol,
+      hostname: url.hostname,
+      ...(url.port && { port: Number(url.port) }),
+      path: url.pathname,
```
tejaskash, line 168: The signed request only carries `url.pathname`, but `post` sends the full `this.endpoint` on line 215. If someone passes a `gatewayEndpoint` (or sets `BEDROCK_AGENTCORE_GATEWAY_ENDPOINT`) with a `?query` on it, the canonical request will not match and the gateway will 403. Real gateway URLs have no query string, so this is an edge case, but it is a confusing one to debug.

`runtime/client.ts:252` handles this by passing `query: Object.fromEntries(url.searchParams.entries())`. Either do that here or reject query strings in `validateEndpointUrl`.

src/_utils/endpoints.ts
```diff
@@ -7,6 +7,56 @@
...
+ *
+ * @param url - The endpoint URL to check
+ * @returns The URL, unchanged
+ *
+ * @throws Error if the URL cannot be parsed or its host is not an AWS host.
+ */
+export function validateEndpointUrl(url: string): string {
+  let hostname: string
+  try {
+    hostname = new URL(url).hostname
```
tejaskash: This accepts `http://something.amazonaws.com/mcp`. The doc comment above says the whole point of the check is that requests get signed with the caller's credentials, and those signed headers carry the access key id and session token. Over plain http they go out in the clear. A `protocol === 'https:'` check would close that and fits what the function already says it is for.

src/tools/web-search/client.ts
```diff
@@ -0,0 +1,776 @@
...
+      throw new Error(`Pass only one of gatewayId, gatewayArn or gatewayEndpoint, got ${given.join(', ')}`)
+    }
+
+    let resolvedId = gatewayId
+    let region = config.region
+
+    if (gatewayArn) {
+      const parsed = parseGatewayArn(gatewayArn)
+      resolvedId = parsed.gatewayId
+      region = region ?? parsed.region
```
tejaskash: When both `region` and `gatewayArn` are given and they disagree, the explicit `region` wins silently (there is a test for this at `client.test.ts:103`). A gateway only lives in one region, so a mismatch is always a caller mistake, and right now it shows up later as a 403 or a DNS failure at search time instead of at construction. I would either throw here when they differ, or ignore `region` when an ARN is given.

src/tools/web-search/client.ts
```diff
@@ -0,0 +1,776 @@
...
+   * `includeDomains` against a target that already has one narrows to the
+   * intersection. Two disjoint include lists return zero results and raise no
+   * error, so check the target's configuration when a filtered search comes back
+   * empty.
+   *
+   * @param query - What to search for, 200 characters or fewer
+   * @param options - Result limit and filters
+   * @returns The search results
+   *
+   * @throws WebSearchError if the call fails or the response cannot be decoded.
```
tejaskash, line 583: The doc says `@throws WebSearchError`, but an empty query or `maxResults: 0` throws a raw `ZodError` out of `buildSearchArguments` on line 532. The runtime does the same thing so this may be on purpose, but then the doc comment should say so. Otherwise wrap the schema failures.

src/tools/web-search/client.ts
```diff
@@ -0,0 +1,776 @@
...
+      this.resetSession()
+      throw error
+    }
+  }
+
+  /**
+   * Resolves the fully qualified tool name, discovering it only when it cannot be
+   * derived, since a `tools/list` costs a round trip.
+   */
+  private async ensureToolName(): Promise<string> {
```
tejaskash, line 340: Unlike `ensureInitialized`, this has no in-flight dedup. If several first searches run concurrently without `targetName`, each one does its own `tools/list` (including any pagination). Not a bug, just wasted round trips. The same promise-sharing trick from `ensureInitialized` would fix it.

src/tools/web-search/client.ts
```diff
@@ -0,0 +1,776 @@
...
+    this.toolName = resolved
+    return resolved
+  }
+
+  /** Lists every tool the gateway exposes, following pagination. */
+  private async listToolNames(): Promise<string[]> {
+    const names: string[] = []
+    let cursor: string | undefined
+
+    for (;;) {
```
tejaskash, line 395: This loop has no upper bound. A gateway that keeps returning the same `nextCursor` would spin forever. A page cap or a check that the cursor actually changed turns that into an error instead of a hang. Only matters against a misbehaving server, so low priority.

src/tools/web-search/client.ts
```diff
@@ -0,0 +1,776 @@
...
+      resolvedId = parsed.gatewayId
+      region = region ?? parsed.region
+    }
+
+    if (!region) {
+      throw new Error('region is required. Pass region, or a gatewayArn to read it from.')
+    }
+
+    if (!(KNOWN_REGIONS as readonly string[]).includes(region)) {
+      console.warn(
```
tejaskash: Nothing else in the library writes to `console.warn` during normal use (the one existing case is the `close()` error path in the Playwright client). This one fires on construction, callers cannot turn it off, and it will start firing wrongly the day a new region launches. I would drop it and leave `KNOWN_REGIONS` exported for callers who want to check themselves, or gate it behind an option.

tejaskash (commented):
Went through the client, the endpoint helper and the tests. The transport code is careful and the SigV4 and credential handling line up with how `runtime/client.ts` and the browser client do it. Nothing here blocks the PR. Inline notes below, roughly ordered from the one I would fix first to the ones that are just cleanup.

src/tools/web-search/client.ts
```diff
@@ -0,0 +1,776 @@
...
+    yield last
+  }
+}
+
+/** Pulls the gateway ID and region out of a gateway ARN. */
+function parseGatewayArn(arn: string): { gatewayId: string; region: string } {
+  const parts = arn.split(':')
+  const resource = parts[5]
+  const region = parts[3]
+  if (parts[0] !== 'arn' || region === undefined || resource === undefined || !resource.startsWith('gateway/')) {
```
tejaskash: This accepts `arn:aws:anything:us-east-1:123:gateway/abc`. Checking `parts[2] === 'bedrock-agentcore'` is cheap and matches what the error message says it expects.

src/tools/web-search/client.ts
sundargthb: Fixed both ways you suggested, since they close different halves of it. `ensureInitialized` now checks `this.initializing` before `this.initialized`, and `initialized` stays false until the notification returns, with `post` taking an explicit `sendProtocolVersion` option so the notification itself still carries the header.

The test starts the second search while the notification is held open and asserts no `tools/call` goes out ahead of it. I checked it fails against the old ordering, since the `Promise.all` version did not.

src/tools/web-search/client.ts
sundargthb: Changed to the SDK version. `src/_utils/version.ts` reads it out of package.json at runtime and exports `SDK_VERSION`, so `scripts/bump-version.ts` keeps it correct on its own. A hardcoded constant would go stale, since that script only rewrites package.json, the lockfile and the changelog.

src/tools/web-search/client.ts
sundargthb: Signed the query rather than rejecting it, the same way `runtime/client.ts:252` does, so a `gatewayEndpoint` carrying one behaves.

The test signs the same initialize request independently with the query included and compares the Authorization header, so removing the query line fails it.

src/_utils/endpoints.ts
sundargthb: Added the `protocol !== 'https:'` check, so `http://something.amazonaws.com/mcp` now throws before anything is signed. The URL is parsed once and both checks run off the parsed value.

src/tools/web-search/client.ts
sundargthb: It throws now, naming both regions. The test that asserted the old precedence is replaced by one for a region that agrees with the ARN and one for a region that disagrees.

This is a deliberate difference from the Python client, which takes `region or arn_region`. A gateway lives in one region, so a mismatch is always a caller mistake, and construction is a better place to say so than a 403 at search time.

src/tools/web-search/client.ts
sundargthb: On purpose, and the doc says so now. A `ZodError` names the field that was wrong and is raised before anything is signed, and wrapping it would lose that. `search()` documents `@throws ZodError` alongside `WebSearchError`, and docs/WEB_SEARCH.md says validation failures arrive as `ZodError` while anything after the request goes out arrives as `WebSearchError`.

src/tools/web-search/client.ts
sundargthb: Added. `ensureToolName` shares a single `resolvingToolName` promise the same way `ensureInitialized` does, with the candidate matching moved into `discoverToolName()`. The test asserts two searches starting together produce one `tools/list` and two tool calls.

src/tools/web-search/client.ts
sundargthb: Bounded both ways. `MAX_TOOL_LIST_PAGES` is 50, and a `nextCursor` equal to the one just sent throws instead of continuing, so a server that repeats a cursor ends the call with an error immediately. Both messages tell the caller to pass `targetName` or `toolName` to skip discovery.

src/tools/web-search/client.ts
sundargthb: Removed. `KNOWN_REGIONS` stays exported and its doc now says it is there for callers who want to check before they call.

The Python client keeps its warning because it goes through `logger.warning`, which an application can silence. `console.warn` cannot be silenced and this repo has no logger to route it through, so dropping it was the better trade here.

src/tools/web-search/client.ts
sundargthb: Added the `parts[2]` check, so `arn:aws:apigateway:us-east-1:123:gateway/abc` throws with the message that already claimed it would. Test covers it.

tejaskash (approved):

PR #267, opened 2026-09-22
fix(runtime): preserve JSON errors before SSE streaming

### LINKED ISSUE #212: BedrockAgentCoreApp throws FST_ERR_REP_INVALID_PAYLOAD_TYPE when a non-streaming handler throws and the client sent Accept: text/event-stream

jariy17:
When an invocation handler throws before streaming and the client accepts SSE, `@fastify/sse` has already staged `Content-Type: text/event-stream`. Fastify then fails to serialize the JSON error, hiding the handler's original message.

Set `application/json` explicitly on the HTTP 500 response while headers are still unsent. The caller receives `{ "error": "downstream dependency failed" }` with HTTP 500; errors after streaming starts continue to use SSE error events.

This retains the current dependency and Node support range. The upstream upgrade proposed in #211 fixes the header timing, but `@fastify/sse` 0.5/0.6 require Node 20.20.2+, above this SDK's declared Node 20.0.0 minimum.

Fixes #212.


jariy17:
Verified commit `a75c1bf8fdfa6ce290e11e5454e1824d67fd0e34`.

- Regression check against the original runtime with `@fastify/sse@0.4.0`: the JSON-error test fails; all three streaming cases pass. With the fix, all four pass on the same dependency.
- Local unit suite: 771 tests passed across 32 files. Lint, source/test type checks, build, and formatting of changed files passed.
- [Integration CI](https://github.com/aws/bedrock-agentcore-sdk-typescript/actions/runs/35797914365): 135 tests passed, 7 skipped, across 8 files.
- Current CI validation, security checks, and CodeQL passed.

Repository-wide formatting still reports the unchanged `.github/workflows/release.yml`; no workflow or dependency files are changed by this PR.

